PR #3, opened 2024-11-04
Populate first draft of color crate

raphlinus:
This is copied from Raph's "koloro" local repository, with lots of fiddly changes to satisfy clippy.

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -1,0 +1,528 @@
+//! Concrete types for colors.
+
+use core::any::TypeId;
+use core::marker::PhantomData;
+
+use crate::{Colorspace, ColorspaceLayout};
+
+#[cfg(all(not(feature = "std"), not(test)))]
+use crate::floatfuncs::FloatFuncs;
+
+/// An opaque color.
+///
+/// A color in a colorspace known at compile time, without transparency. Note
+/// that "opaque" refers to the color, not the representation; the components
+/// are publicly accessible.
+#[derive(Clone, Copy, Debug)]
+pub struct OpaqueColor<T> {
+    pub components: [f32; 3],
+    pub cs: PhantomData<T>,
+}
+
+/// A color with an alpha channel.
+///
+/// A color in a colorspace known at compile time, with an alpha channel.
+#[derive(Clone, Copy, Debug)]
+pub struct AlphaColor<T> {
+    pub components: [f32; 4],
+    pub cs: PhantomData<T>,
+}
+
+/// A color with premultiplied alpha.
+///
+/// A color in a colorspace known at compile time, with a premultiplied
+/// alpha channel.
+///
+/// Following the convention of CSS Color 4, in cylindrical color spaces
+/// the hue channel is not premultiplied. If it were, interpolation would
+/// give undesirable results.
+#[derive(Clone, Copy, Debug)]
+pub struct PremulColor<T> {
+    pub components: [f32; 4],
+    pub cs: PhantomData<T>,
+}
+
+/// The hue direction for interpolation.
+///
+/// This type corresponds to `hue-interpolation-method` in the CSS Color
+/// 4 spec.
+#[derive(Clone, Copy, Default, Debug)]
+#[non_exhaustive]
+pub enum HueDirection {
+    #[default]
+    Shorter,
+    Longer,
+    Increasing,
+    Decreasing,
+    // It's possible we'll add "raw"; color.js has it.
+}
+
+/// Fixup hue based on specified hue direction.
+///
+/// Reference: §12.4 of CSS Color 4 spec
+///
+/// Note that this technique has been tweaked to only modify the second hue.
+/// The rationale for this is to support multiple gradient stops, for example
+/// in a spline. Apply the fixup to successive adjacent pairs.
+///
+/// In addition, hues outside [0, 360) are supported, with a resulting hue
+/// difference always in [-360, 360].
+fn fixup_hue(h1: f32, h2: &mut f32, direction: HueDirection) {
+    let dh = (*h2 - h1) * (1. / 360.);
+    match direction {
+        HueDirection::Shorter => {
+            // Round, resolving ties toward zero.
+            let rounded = if dh - dh.floor() == 0.5 {
+                dh.trunc()
+            } else {
+                dh.round()
+            };
+            *h2 -= 360. * rounded;
+        }
+        HueDirection::Longer => {
+            let t = 2.0 * dh.abs().ceil() - (dh.abs() + 1.5).floor();
+            *h2 += 360.0 * (t.copysign(0.0 - dh));
+        }
+        HueDirection::Increasing => *h2 -= 360.0 * dh.floor(),
+        HueDirection::Decreasing => *h2 -= 360.0 * dh.ceil(),
+    }
+}
+
+pub(crate) fn fixup_hues_for_interpolate(
+    a: [f32; 3],
+    b: &mut [f32; 3],
+    layout: ColorspaceLayout,
+    direction: HueDirection,
+) {
+    if let Some(ix) = layout.hue_channel() {
+        fixup_hue(a[ix], &mut b[ix], direction);
+    }
+}
+
+impl<T: Colorspace> OpaqueColor<T> {
+    pub const fn new(components: [f32; 3]) -> Self {
+        let cs = PhantomData;
+        Self { components, cs }
+    }
+
+    pub fn convert<U: Colorspace>(self) -> OpaqueColor<U> {
+        if TypeId::of::<T>() == TypeId::of::<U>() {
+            OpaqueColor::new(self.components)
+        } else {
+            let lin_rgb = T::to_linear_srgb(self.components);
+            OpaqueColor::new(U::from_linear_srgb(lin_rgb))
+        }
+    }
+
+    /// Add an alpha channel.
+    ///
+    /// This function is the inverse of [`AlphaColor::split`].
+    pub const fn with_alpha(self, alpha: f32) -> AlphaColor<T> {
+        AlphaColor::new(add_alpha(self.components, alpha))
+    }
+
+    /// Difference between two colors by Euclidean metric.
+    pub fn difference(self, other: Self) -> f32 {
+        let x = self.components;
+        let y = other.components;
+        let (d0, d1, d2) = (x[0] - y[0], x[1] - y[1], x[2] - y[2]);
+        (d0 * d0 + d1 * d1 + d2 * d2).sqrt()
+    }
+
+    /// Linearly interpolate colors, without hue fixup.
+    ///
+    /// This method produces meaningful results in rectangular colorspaces,
+    /// or if hue fixup has been applied.
+    #[must_use]
+    pub fn lerp_rect(self, other: Self, t: f32) -> Self {
+        self + t * (other - self)
+    }
+
+    /// Apply hue fixup for interpolation.
+    ///
+    /// Adjust the hue angle of `other` so that linear interpolation results in
+    /// the expected hue direction.
+    pub fn fixup_hues(self, other: &mut Self, direction: HueDirection) {
+        fixup_hues_for_interpolate(self.components, &mut other.components, T::LAYOUT, direction);
+    }
+
+    /// Linearly interpolate colors, with hue fixup if needed.
+    #[must_use]
+    pub fn lerp(self, mut other: Self, t: f32, direction: HueDirection) -> Self {
+        self.fixup_hues(&mut other, direction);
+        self.lerp_rect(other, t)
+    }
+
+    /// Scale the chroma by the given amount.
+    ///
+    /// See [`Colorspace::scale_chroma`] for more details.
+    #[must_use]
+    pub fn scale_chroma(self, scale: f32) -> Self {
+        Self::new(T::scale_chroma(self.components, scale))
+    }
+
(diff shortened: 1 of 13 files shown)


raphlinus:
A few more things to say about this. A few things are still in flux, though overall I'm happy with the direction it's going. I am especially curious to learn whether it meets the needs of potential users, or has some gaps.

Not all the CSS color spaces are done. Adding those should be fairly straightforward.

This crate would benefit from extensive testing. There are a bunch of tests in the CSS Color 4 spec, and likely many of those could be adapted without much trouble. In general, the [color.js](https://github.com/color-js/color.js) library can be considered a source of truth, and perhaps tests could be written to test whether this library gives consistent results (I'm thinking of the extremely extensive fuzz testing against the reference cmark implementation in pulldown-cmark, which also tests against a JS ref).

The `TaggedColor` and `CssColor` types should almost certainly be merged, I don't think there's significant value in having them separate. The main difference is that the former doesn't have the concept of missing channels. I think the runtime cost of supporting them is minimal.

There are a few other todos and discussion questions in the code.

Discussion is welcome! There's more than a bit of background in a [Zulip thread](https://xi.zulipchat.com/#narrow/channel/260979-kurbo/topic/Peniko.20interpolate.20between.20Colors/near/479534212).

color/src/color.rs
```diff
@@ -0,0 +1,531 @@
...
+impl<T: Colorspace> core::ops::Div<f32> for OpaqueColor<T> {
+    type Output = Self;
+
+    #[expect(
+        clippy::suspicious_arithmetic_impl,
+        reason = "somebody please teach clippy about multiplicative inverses"
+    )]
+    fn div(self, rhs: f32) -> Self {
+        self * rhs.recip()
+    }
```
xStrom, line 360: 😄

xStrom (approved):
I did a quick pass and didn't find any obvious faults. Did not verify magic numbers or color theory.

Given the fresh nature of the repo, I think it's fine to merge and do improvements in follow-up PRs.

DJMcNab (approved):
I tend to agree that landing is probably reasonable.
I do intend to do a proper code style review pass tomorrow.

The one other thing to check is whether the AUTHORS file matches your expectations

color/src/color.rs
```diff
@@ -0,0 +1,531 @@
...
+    ///
+    /// This can be useful for choosing contrasting colors, and follows the
+    /// WCAG 2.1 spec.
+    pub fn relative_luminance(self) -> f32 {
+        let rgb = T::to_linear_srgb(self.components);
+        0.2126 * rgb[0] + 0.7152 * rgb[1] + 0.0722 * rgb[2]
+    }
+}
+
+pub(crate) const fn split_alpha(x: [f32; 4]) -> ([f32; 3], f32) {
```
DJMcNab, line 177: This could potentially use `[x, y, z, a]: [f32; 4]`. This should only be a code-style thing.

color/src/color.rs
```diff
@@ -0,0 +1,531 @@
...
+    pub fn un_premultiply(self) -> AlphaColor<T> {
+        let (multiplied, alpha) = split_alpha(self.components);
+        let scale = if alpha == 0.0 { 1.0 } else { 1.0 / alpha };
+        AlphaColor::new(add_alpha(T::LAYOUT.scale(multiplied, scale), alpha))
+    }
+
+    /// Interpolate colors.
+    ///
+    /// Note: this function doesn't fix up hue in cylindrical spaces. It is
+    /// still be useful if the hue angles are compatible, particularly if
```
DJMcNab, line 276: It is still ~~be~~ useful

color/src/color.rs
```diff
@@ -0,0 +1,531 @@
...
+}
+
+impl<T: Colorspace> From<OpaqueColor<T>> for PremulColor<T> {
+    fn from(value: OpaqueColor<T>) -> Self {
+        Self::new(add_alpha(value.components, 1.0))
+    }
+}
+
+// Arithmetic traits. A major motivation for providing these is to enable
+// weighted sums, which are well defined when the weights sum to 1. In
```
DJMcNab, line 331: I don't have the domain knowledge here, but this sentence *suggests* that the operations implemented here are potentially suspect, and might be better as methods with documentation and caveats.

At the very least, the impls themselves probably should have documentation explaining some of this.

color/src/color.rs
```diff
@@ -0,0 +1,531 @@
...
+        rhs * self
+    }
+}
+
+impl<T: Colorspace> core::ops::Div<f32> for OpaqueColor<T> {
+    type Output = Self;
+
+    #[expect(
+        clippy::suspicious_arithmetic_impl,
+        reason = "somebody please teach clippy about multiplicative inverses"
```
DJMcNab, line 356: Is it worth linking to https://github.com/rust-lang/rust-clippy/issues/13652 here?

color/src/color.rs
```diff
@@ -0,0 +1,531 @@
...
+        Self::new([x[0] - y[0], x[1] - y[1], x[2] - y[2], x[3] - y[3]])
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::{fixup_hue, HueDirection};
+
+    #[test]
+    fn test_hue_fixup() {
```
DJMcNab, line 484: I think that this `test_` prefix is unidiomatic. See e.g. https://github.com/rust-lang/rust-clippy/issues/8931

color/src/color.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    fn sub(self, rhs: Self) -> Self {
+        let x = self.components;
+        let y = rhs.components;
+        Self::new([x[0] - y[0], x[1] - y[1], x[2] - y[2], x[3] - y[3]])
+    }
+}
+
+#[cfg(test)]
+mod test {
```
DJMcNab, line 480: Similarly, this would normally be called `tests`.

color/src/color.rs
```diff
@@ -0,0 +1,531 @@
...
+                let dh = h2 - h1;
+                let mut fixed_h2 = h2;
+                fixup_hue(h1, &mut fixed_h2, HueDirection::Shorter);
+                let (mut spec_h1, mut spec_h2) = (h1, h2);
+                if dh > 180.0 {
+                    spec_h1 += 360.0;
+                } else if dh < -180.0 {
+                    spec_h2 += 360.0;
+                }
+                assert_eq!(fixed_h2 - h1, spec_h2 - spec_h1);
```
DJMcNab, line 498: If I were writing this test, I'd probably put these all in their own "anonymous" scope, so that there's no risk of forgetting to reset a variable by accident.

color/src/colorspace.rs
```diff
@@ -0,0 +1,287 @@
...
+///
+/// The linear sRGB colorspace is central, and other colorspaces are
+/// defined as conversions in and out of that. A colorspace does not
+/// explicitly define a gamut, so generally conversions will succeed
+/// and round-trip, subject to numerical precision.
+///
+/// White point is not explicitly represented. For color spaces with a
+/// white point other than D65 (the native white point for sRGB), use
+/// a linear Bradford chromatic adaptation, following CSS Color 4.
+pub trait Colorspace: Clone + Copy + 'static {
```
DJMcNab, line 24: I'd *probably* expect this to be `ColorSpace`; the css color spec uses the term "colorspace" 9 times, and the term "color space" ~200 times (of note, it also uses the term `color-space` ~9 times)

color/src/colorspace.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// Whether the colorspace is linear.
+    ///
+    /// Calculations in linear colorspaces can sometimes be simplified,
+    /// for example it is not necessary to undo premultiplication when
+    /// converting.
+    const IS_LINEAR: bool = false;
```
DJMcNab, line 30: Is having defaults here good?

color/src/colorspace.rs
```diff
@@ -0,0 +1,287 @@
...
+            [0.019_330_818, 0.119_194_78, 0.950_532_14],
+        ];
+        matmul(&LINEAR_SRGB_TO_XYZ, src)
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct Oklab;
+
+// Matrices taken from Oklab blog post, precision reduced to f32
```
DJMcNab, line 214: Link?

color/src/colorspace.rs
```diff
@@ -0,0 +1,287 @@
...
+
+#[derive(Clone, Copy, Debug)]
+pub struct Oklch;
+
+impl Colorspace for Oklch {
+    const CS_TAG: Option<ColorspaceTag> = Some(ColorspaceTag::Oklch);
+
+    const LAYOUT: ColorspaceLayout = ColorspaceLayout::HueThird;
+
+    fn from_linear_srgb(src: [f32; 3]) -> [f32; 3] {
```
DJMcNab, line 265: In `bevy_color`, I believe it is possible to convert between `Oklab` and `Oklch` directly. Would this be a useful operation to support?

color/src/css.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub fn relative_luminance(self) -> f32 {
+        let rgb = self.convert(ColorspaceTag::LinearSrgb).components;
+        0.2126 * rgb[0] + 0.7152 * rgb[1] + 0.0722 * rgb[2]
+    }
+}
+
+impl Interpolator {
+    pub fn eval(&self, t: f32) -> CssColor {
+        let premul = [
+            self.premul1[0] + t * (self.premul2[0] - self.premul1[0]),
```
DJMcNab, line 205: It doesn't look like `premul2` is ever used except to compute the same difference value here.

Should this difference instead be stored?

color/src/floatfuncs.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright 2024 the Color Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Shims for math functions that ordinarily come from std.
+
+/// Defines a trait that chooses between libstd or libm implementations of float methods.
+macro_rules! define_float_funcs {
+    ($(
+        fn $name:ident(self $(,$arg:ident: $arg_ty:ty)*) -> $ret:ty
+        => $lname:ident/$lfname:ident;
```
DJMcNab, line 10: `lname` doesn't seem to be used?

color/src/gradient.rs
```diff
@@ -0,0 +1,89 @@
...
+    tolerance: f32,
+    // The adaptive subdivision logic is lifted from the stroke expansion paper.
+    t0: u32,
+    dt: f32,
+    target0: PremulColor<CS>,
+    target1: PremulColor<CS>,
+    end_color: PremulColor<CS>,
+}
+
+pub fn gradient<CS: Colorspace>(
```
DJMcNab, line 19: More evidence in favour of `ColorSpace` in this generic name

DJMcNab (commented):
A single-pass review sweep.

color/src/color.rs
raphlinus: I've thought about this deeply and feel that arithmetic operations are worthwhile to include. Basically, it enables spline interpolation with a generic spline implementation that has the arithmetic traits as bounds, and to me that upside outweighs any downside. Other color crates impl these as well.

I have moved this comment to documentation.

color/src/colorspace.rs
raphlinus: For `IS_LINEAR`, yes, for two reasons. First, I find myself motivated by code size, and most colorspaces are not linear. Second, it is safe; if people forget to set it, the only thing that goes wrong is a bit of extra conversion code.

For `CS_TAG`, yes, because it will always be `None` in client impls.

The argument for `LAYOUT` is less compelling, but if we already have defaults I think it makes sense to go with the flow.

color/src/colorspace.rs
raphlinus: I was wrestling between code size and runtime efficiency. But on reflection, I think it makes sense to have a `convert<TargetCS: ColorSpace>` method with a default impl. Probably I'll do that in a separate PR to keep the next one focused on cleanup.

color/src/css.rs
raphlinus: As a bit of super-geekery, there are two basic strategies for implementing lerp. One is `a + t * (b - a)`, the other is `a * (1 - t) + b * t`. The latter has the advantage of zero rounding error at t = 1, but I think for the purposes of this crate one ulp at f32 is completely acceptable.

color/src/floatfuncs.rs
raphlinus: Yup, these were adapted from kurbo, but I think we can simplify and only include f32 versions.

color/src/css.rs
waywardmonkeys: `glam` ended up switching to the precise lerp (reported in https://github.com/bitshifter/glam-rs/issues/275, fixed in https://github.com/bitshifter/glam-rs/pull/548 (but some of the concerns about differences in magnitude are different in color code vs generic vectors ...)

color/src/color.rs
DJMcNab: Seems reasonable. An alternative curve bending design might be to have an `Arithmetic<T>` wrapper, where you're opting in to thinking carefully about this.

color/src/css.rs
raphlinus: Right, for geometry sometimes you really care about watertightness, but for color close is pretty much always good enough.

PR #6, opened 2024-11-05
Add color clipping

tomcur:
Clipping colors to colorspaces' bounds is useful for e.g. final display or as part of gamut mapping.

A related method would be `Colorspace::is_in_bounds(src: [f32; 3])`, but I'm undecided whether that's useful enough to include. It could have a default implementation (`src == Self::clip(src)`). Roughly the same considerations hold for a const `Colorspace::IS_(UN)BOUNDED`.

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -75,6 +75,21 @@ pub trait ColorSpace: Clone + Copy + 'static {
             TargetCS::from_linear_srgb(lin_rgb)
         }
     }
+
+    /// Clip the color's components to the range allowed by the colorspace.
+    ///
+    /// The resultant color is guaranteed to be inside the bounds (and thus gamut) of the
+    /// colorspace, but may be perceptually quite distinct from the original color.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use color::{Colorspace, Srgb, XyzD65};
+    ///
+    /// assert_eq!(Srgb::clip([0.4, -0.2, 1.2]), [0.4, 0., 1.]);
+    /// assert_eq!(XyzD65::clip([0.4, -0.2, 1.2]), [0.4, -0.2, 1.2]);
+    /// ```
+    fn clip(src: [f32; 3]) -> [f32; 3];
 }
 
 /// The layout of a color space, particularly the hue channel.
@@ -139,6 +154,14 @@ impl ColorSpace for LinearSrgb {
         ];
         matmul(&OKLAB_LMS_TO_SRGB, lms_scaled.map(|x| x * x * x))
     }
+
+    fn clip(src: [f32; 3]) -> [f32; 3] {
+        [
+            src[0].clamp(0., 1.),
+            src[1].clamp(0., 1.),
+            src[2].clamp(0., 1.),
+        ]
+    }
 }
 
 // It might be a better idea to write custom debug impls for AlphaColor and friends
@@ -171,6 +194,14 @@ impl ColorSpace for Srgb {
     fn from_linear_srgb(src: [f32; 3]) -> [f32; 3] {
         src.map(lin_to_srgb)
     }
+
+    fn clip(src: [f32; 3]) -> [f32; 3] {
+        [
+            src[0].clamp(0., 1.),
+            src[1].clamp(0., 1.),
+            src[2].clamp(0., 1.),
+        ]
+    }
 }
 
 #[derive(Clone, Copy, Debug)]
@@ -195,6 +226,14 @@ impl ColorSpace for DisplayP3 {
             [0.017_082_632, 0.072_397_44, 0.910_519_96],
         ];
         matmul(&LINEAR_SRGB_TO_DISPLAYP3, src).map(lin_to_srgb)
+    }
+
+    fn clip(src: [f32; 3]) -> [f32; 3] {
+        [
+            src[0].clamp(0., 1.),
+            src[1].clamp(0., 1.),
+            src[2].clamp(0., 1.),
+        ]
     }
 }
 
@@ -222,6 +261,10 @@ impl ColorSpace for XyzD65 {
             [0.019_330_818, 0.119_194_78, 0.950_532_14],
         ];
         matmul(&LINEAR_SRGB_TO_XYZ, src)
+    }
+
+    fn clip(src: [f32; 3]) -> [f32; 3] {
+        src
     }
 }
 
@@ -282,6 +325,10 @@ impl ColorSpace for Oklab {
             TargetCS::from_linear_srgb(lin_rgb)
         }
     }
+
+    fn clip(src: [f32; 3]) -> [f32; 3] {
+        [src[0].clamp(0., 1.), src[1], src[2]]
+    }
 }
 
 /// Rectangular to cylindrical conversion.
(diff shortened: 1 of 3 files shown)


color/src/css.rs
```diff
@@ -124,6 +124,20 @@ impl CssColor {
...
+    pub fn clip(self) -> Self {
+        let (opaque, alpha) = split_alpha(self.components);
+        let components = self.cs.clip(opaque);
+        Self {
+            cs: self.cs,
+            missing: self.missing,
+            components: add_alpha(components, alpha),
+        }
+    }
+
```
tomcur, line 140: Alpha is unchanged here, maybe it should be clamped to [0, 1].

color/src/colorspace.rs
```diff
@@ -252,6 +295,10 @@ impl Colorspace for Oklab {
     fn scale_chroma(src: [f32; 3], scale: f32) -> [f32; 3] {
         [src[0], src[1] * scale, src[2] * scale]
     }
+
+    fn clip(src: [f32; 3]) -> [f32; 3] {
```
raphlinus, line 299: Note the style point from #7, this could be `[l, a, b]: [f32; 3]`. I *didn't* apply that completely consistently (see scale_chroma above) but might be a good idea going forward.

color/src/css.rs
raphlinus: I think so, yes.

Also a really subtle point I'll raise here, that doesn't need change in the code. There is a not-strictly-enforced invariant that missing channels have 0.0 components. Since clipping will always preserve 0, there's no fixup needed here.

color/src/colorspace.rs
```diff
@@ -58,6 +58,21 @@ pub trait Colorspace: Clone + Copy + 'static {
         let scaled = LinearSrgb::scale_chroma(rgb, scale);
         Self::from_linear_srgb(scaled)
     }
+
+    /// Clip the color's components to the range allowed by the colorspace.
```
raphlinus, line 62: I really don't like "allowed by the color space," as my intent is for color spaces to allow all colors. I think I'd phrase this, "to fit within the natural gamut of the color space."

color/src/colorspace.rs
```diff
@@ -58,6 +58,21 @@ pub trait Colorspace: Clone + Copy + 'static {
...
+    ///
+    /// The resultant color is guaranteed to be inside the bounds (and thus gamut) of the
```
raphlinus, line 64: I think what I'd say here is that there are many possible mappings from out-of-gamut colors to the gamut boundary. Some may be perceptually better than others, for example preserving hue. This method will generally do the mathematically simplest thing, namely clamping of individual component values.

raphlinus (approved):
I think this can go in (modulo rebasing/renaming as per #7 and the language change suggested).

color/src/colorspace.rs
tomcur: Agreed, especially seeing the documentation on `ColorSpace` that it "does not explicitly define a gamut". I changed the language

color/src/css.rs
tomcur: The alpha is now also clamped.

I did not copy the set-missing-to-0 from the `scale_chroma` function above for that reason; it may be worth documenting the invariant itself.

color/src/colorspace.rs
```diff
@@ -206,6 +245,10 @@ impl Colorspace for XyzD65 {
         ];
         matmul(&LINEAR_SRGB_TO_XYZ, src)
     }
+
+    fn clip(src: [f32; 3]) -> [f32; 3] {
+        src
```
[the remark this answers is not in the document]
tomcur: I'm not sure about clamping XYZ, e.g., changing the reference illumination by going from XYZ D65 to D50 negative values often roll out of the maths (but only, I believe, for quite extreme colors). I don't know whether those colors are physically possible.

color/src/css.rs
```diff
@@ -127,6 +127,22 @@ impl CssColor {
...
 
+    /// Clip the color's components to fit within the natural gamut of the color space, and clamp
+    /// the color's alpha to be in the range `[0, 1]`.
+    ///
+    /// See [`ColorSpace::clip`] for more details.
+    #[must_use]
+    pub fn clip(self) -> Self {
+        let (opaque, alpha) = split_alpha(self.components);
+        let components = self.cs.clip(opaque);
+        let alpha = alpha.clamp(0., 1.);
```
DJMcNab, line 138: Would it be meaningful to have a separate `clip_alpha` method? Semantically, a less than zero alpha is definitely equivalent to zero, and likely similarly for greater than 1?

color/src/tagged.rs
```diff
@@ -178,6 +178,21 @@ impl ColorSpaceTag {
...
+    /// See [`ColorSpace::clip`] for more details.
+    pub fn clip(self, src: [f32; 3]) -> [f32; 3] {
+        match self {
+            Self::Srgb => Srgb::clip(src),
+            Self::LinearSrgb => LinearSrgb::clip(src),
+            Self::Oklab => Oklab::clip(src),
+            Self::Oklch => Oklch::clip(src),
+            Self::DisplayP3 => DisplayP3::clip(src),
+            Self::XyzD65 => XyzD65::clip(src),
+            _ => todo!(),
```
DJMcNab, line 193: This isn't ideal, but I think this is *fine*. The expectation is that all three `todo!`s in this file would be fixed as a block

DJMcNab (commented):
(Prior approval still stands, and you should merge if this is ready)

I have not validated the logic, just some drive-by thoughts.

color/src/colorspace.rs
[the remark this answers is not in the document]
tomcur: I've merged the PR – a follow-up could be done either commenting `clip` for XYZ, or, more generally, what the color space components' represent for each color space (and, in particular, what their natural bounds are).

color/src/tagged.rs
tomcur: Some tags are not yet implemented. To move the concerns (and code duplication) to a single place, perhaps a private `match`-like macro that takes by-name a fallback `ColorSpace` method makes sense for the ergonomics. `ColorSpace` is not currently object-safe so we can't go through a `&dyn ColorSpace`.

color/src/css.rs
tomcur: Maybe; would `clip` leave alpha unchanged in that case?

color/src/tagged.rs
tomcur: (Alternatively, we deal with this when more color spaces are implemented and accept the code duplication.)

raphlinus: We will probably end up with a macro, but keep in mind some methods do really want custom logic; `convert` is the big one.

color/src/css.rs
DJMcNab: I don't think so? The other thing to think about is clipping the alpha for premultiplied

tomcur: I think that makes a good case to add it as a convenience method. Clipping unconditionally to premultiply would be a performance hit (plus premul math still works out when out-of-bounds, if `alpha` is interpreted as an arbitrary weighting).

color/src/tagged.rs
DJMcNab: I suspect for the amount of things, a macro is probably not necessary. But we can discuss that when the last implementations land.

raphlinus:
I'll put this comment here. I'm pretty sure negative XYZ values don't occur in the normal course of events. Of course negative RGB values will occur in any RGB-based color space, for values in the chromaticity horseshoe but outside the RGB triangle. But I think it's basically an academic question, as clipping in XYZ is also seldom useful.

PR #7, opened 2024-11-05
Cleanups and renames

raphlinus:
In response to feedback on #3.

One comment that was not addressed was direct conversion between color spaces, as opposed to always going through linear rgb.

### DIFF
--- a/color/src/tagged.rs
+++ b/color/src/tagged.rs
@@ -9,7 +9,7 @@ use crate::{
     XyzD65,
 };
 
-/// The colo rspace tag for tagged colors.
+/// The color space tag for tagged colors.
 ///
 /// This represents a fixed set of known color spaces. The set is
 /// based on the CSS Color 4 spec, but might also extend to a small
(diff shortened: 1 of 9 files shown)


color/src/tagged.rs
```diff
@@ -1,23 +1,23 @@
...
 
 use crate::{
     color::{add_alpha, split_alpha},
-    AlphaColor, Bitset, Colorspace, ColorspaceLayout, DisplayP3, LinearSrgb, Oklab, Oklch, Srgb,
+    AlphaColor, Bitset, ColorSpace, ColorSpaceLayout, DisplayP3, LinearSrgb, Oklab, Oklch, Srgb,
     XyzD65,
 };
 
-/// The colorspace tag for tagged colors.
+/// The colo rspace tag for tagged colors.
```
waywardmonkeys, line 12: Space in wrong place!

DJMcNab (approved):
Looks good. One more space misplacement

PR #8, opened 2024-11-05
Move conversion logic into ColorSpace

raphlinus:
Add a `convert` method to `ColorSpace`, and migrate logic in the individual color types to use that.

One consequence is that gradient interpolation in Oklch is expected to be faster, as the conversion to Oklab for error evaluation now doesn't go all the way back to linear sRGB.

Also adds examples from koloro crate, which were inadvertently omitted from the initial commit.

Removes `TaggedColor::from_linear_srgb` method, as that functionality can now be expressed more idiomatically with properly typed colors.

### DIFF
--- a/color/src/tagged.rs
+++ b/color/src/tagged.rs
@@ -145,6 +145,15 @@ impl ColorSpaceTag {
         }
     }
 
+    pub fn convert(self, target: Self, src: [f32; 3]) -> [f32; 3] {
+        match (self, target) {
+            _ if self == target => src,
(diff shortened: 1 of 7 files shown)


color/src/tagged.rs
```diff
@@ -145,6 +145,15 @@ impl ColorSpaceTag {
         }
     }
 
+    pub fn convert(self, target: Self, src: [f32; 3]) -> [f32; 3] {
```
DJMcNab, line 148: Maybe this could have a short doc comment?

DJMcNab (approved):
Looks good.

PR #9, opened 2024-11-06
Add methods to map color components

raphlinus:
These are convenience methods, but should provide a nice concise way to express common operations on colors, subsuming a lot of single-use methods.

Arguably 'map_lightness' should become a method on `ColorSpace`, very similar to `scale_chroma`, as similar math optimizations apply.

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -6,7 +6,7 @@
 use core::any::TypeId;
 use core::marker::PhantomData;
 
-use crate::{ColorSpace, ColorSpaceLayout};
+use crate::{ColorSpace, ColorSpaceLayout, ColorSpaceTag, Oklab};
 
 #[cfg(all(not(feature = "std"), not(test)))]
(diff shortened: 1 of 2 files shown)


xorgy (approved):

color/src/color.rs
```diff
@@ -179,9 +179,40 @@ impl<CS: ColorSpace> OpaqueColor<CS> {
...
+        self.convert::<TargetCS>().map(f).convert()
+    }
+
+    /// Map the lightness of the color.
+    ///
+    /// In a color space that naturally has a lightness component, map that value.
+    /// Otherwise, do the mapping in Oklab. The lightness range is normalized so
+    /// that 1.0 is white.
+    #[must_use]
+    pub fn map_lightness(self, f: impl Fn(f32) -> f32) -> Self {
```
tomcur, line 207: Another candidate would be `map_hue` due to its natural representation in all polar spaces.

tomcur (approved):

PR #12, opened 2024-11-07
Add x11 named colors

raphlinus:
Add a lookup for x11 named colors to the parser. The implementation is a hand-rolled minimal perfect hash.


raphlinus:
Some benchmark results, this is parsing "lightgoldenrodyellow" and "red" respectively.

```
test parse_lgy ... bench:          25.42 ns/iter (+/- 30.12)
test parse_red ... bench:          24.06 ns/iter (+/- 0.21)
```

Cargo asm reports 60 lines of asm for color::x11_colors::lookup_palette, which as a rough estimate is 240 bytes (a nontrivial amount of time was spent in Godbolt shaving this down). I expect ballpark 3.6k for the data (142 * 16 bytes for the pointers + 142 * 1 byte for the salts + 1.2k for the name strings).

Here's peniko:

```
test parse_lgy ... bench:          17.99 ns/iter (+/- 0.49)
test parse_red ... bench:          29.15 ns/iter (+/- 0.17)
```

I believe the "red" takes longer because it's farther down the alphabet. Cargo bloat reports 8.1KiB for `peniko::color::parse_color`, while cargo asm reports 2536 lines. That includes the hex parsing, but the color names are the lion's share (most of the asm is highly unrolled codegen of the match statement, loading ASCII string chunks into registers for comparison).

Here's svgtypes:

```
test parse_lgy ... bench:          66.13 ns/iter (+/- 2.50)
test parse_red ... bench:          49.77 ns/iter (+/- 1.19)
```

Here a large part of the cost is [siphasher](https://docs.rs/siphasher/latest/siphasher/), as can be seen by the string length dependency. Cargo bloat reports 296 bytes for `svgtypes::colors::get_index` but I expect that's an underestimate because I don't think it includes siphasher. The cost of the data is expected to be basically the same as this PR.

There is an opportunity to shave ~2k off the data, by concatenating all the strings and storing the offset and length into that, packed into a u16 (5 bits for length, 11 for offset, which fits exactly). I'm not sure that's worth the hassle at this point, but if we get pressure to reduce binary size, it's doable.

waywardmonkeys:
Will there be something like a module `named` which allows for referring to them directly by name, like `color::named::RED` or something similar? Peniko has `peniko::Color::RED` and `bevy_color` has `bevy_color::palettes::basic:RED` and similar for css and tailwind.

raphlinus:
Yes, it's a valid question, just a question of bikeshedding the module structure really. Such a module would have no impact on code size and minimal on compile time. Supporting that is one reason `AlphaColor<Srgb>::from_rgba8()` is const.

color/make_x11_colors.py
```diff
@@ -0,0 +1,269 @@
...
+    ("slateblue", (106, 90, 205, 255)),
+    ("slategray", (112, 128, 144, 255)),
+    ("snow", (255, 250, 250, 255)),
+    ("springgreen", (0, 255, 127, 255)),
+    ("steelblue", (70, 130, 180, 255)),
+    ("tan", (210, 180, 140, 255)),
+    ("teal", (0, 128, 128, 255)),
+    ("thistle", (216, 191, 216, 255)),
+    ("tomato", (255, 99, 71, 255)),
+    ("transparent", (0, 0, 0, 0)),
```
tomcur, line 148: `transparent` is a special-case in `parse::parse_color`. If that remains, this row can be dropped and then alpha does not need to be stored.

https://github.com/linebender/color/blob/e824f679228476722275a25d2107635743b5a33e/color/src/parse.rs#L376-L383

waywardmonkeys: I think this same list will be used to generate a bunch of const values like `color::named::TRANSPARENT` or whatever, and then we would want it here.

DJMcNab: Special casing `TRANSPARENT` in that case would also be fine IMO, so shouldn't be a blocker for this suggestion. Of note, [`bevy_color`](https://docs.rs/bevy_color/latest/bevy_color/palettes/index.html) doesn't provide the transparent colour.

color/make_x11_colors.py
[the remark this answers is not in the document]
Philipp-M: One of the rare cases where AI can help (as code-quality is not of high concern here, and conversion of short scripts from one to another language IME works *ok*)?

color/make_x11_colors.py
raphlinus: I've deliberately moved it out, partly as a performance optimization, partly because transparent is an expected palette color. I suppose it could save 142 bytes, as the 3-byte colors would be stored packed, but I'm not sure it's worth it.

raphlinus:
For the color constants, I propose `color::css_palette::RED` and so on. Is that something that should go into this PR, or a separate one?

Also a note, related to this functionality. The rules for CSS serialization require preserving the name of the color for these named colors. That is useful when doing things like CSS processing, where the output goes to the web and should be human readable, but superfluous in all other cases such as rendering. If we wanted to do it, we could - use one bit in the bitset, and either use linear search or do another phf for the inverse map (I think the latter can be done in 284 bytes of extra data, reusing the existing color data and names).

waywardmonkeys:
bikeshedding, but I prefer `color::palette::css::RED` over `color::css_palette::RED` as I was thinking that `use color::palette::css;` and `css::RED` is nicer than `use color::css_palette;` and `css_palette::RED`.

I was going to suggest having other palettes available like `bevy_color` does with `tailwind`, but then I talked myself out of it and wanting something better that we can talk about in the future.

waywardmonkeys:
>  If we wanted to do it, we could - use one bit in the bitset, and either use linear search or do another phf for the inverse map (I think the latter can be done in 284 bytes of extra data, reusing the existing color data and names).

If a color has gone through a conversion or two, would we have numerical issues with getting back the name? (Would that matter?)

raphlinus:
> If a color has gone through a conversion or two, would we have numerical issues with getting back the name? (Would that matter?)

My sense here is that if the name bit is set, then you round to 8 bit and look up the name after rounding. That should settle any such questions. Of course if there's not a match after that, you serialize based on the numeric values.

waywardmonkeys:
> > If a color has gone through a conversion or two, would we have numerical issues with getting back the name? (Would that matter?)
> 
> My sense here is that if the name bit is set, then you round to 8 bit and look up the name after rounding. That should settle any such questions. Of course if there's not a match after that, you serialize based on the numeric values.

👍

waywardmonkeys (approved):
Constants and serialization can be follow-ups, I think.

PR #25, opened 2024-11-09
Merge TaggedColor and CssColor

### LINKED ISSUE #17: Consider renaming `Bitset`
This is only used for `missing` and doesn't need to be any more general purpose than what's needed there.

Consider renaming this to make it clear that it only exists for one thing and isn't meant to be a general purpose utility struct.

### LINKED ISSUE #21: Merge `CSSColor` and `TaggedColor`
Raph leans towards keeping the `TaggedColor` name but with merged functionality. (I think.)

raphlinus:
As per discussion, the new name is `DynamicColor`, following the precedent of the `image` crate.

Also do some renames and add some docstrings.

Closes #21. The rename of Bitset closes #17.

### DIFF
--- a/color/src/missing.rs
+++ b/color/src/missing.rs
@@ -19,7 +19,7 @@ impl Missing {
     }
 
     /// The set containing a single component index.
-    pub fn singleton(ix: usize) -> Self {
+    pub fn single(ix: usize) -> Self {
         Self(1 << ix)
     }
 
--- a/color/src/missing.rs
+++ b/color/src/missing.rs
@@ -3,29 +3,33 @@
 
 //! A simple bitset.
 
-/// A simple bitset, for representing missing components.
+/// A simple bitset for representing missing components.
 #[derive(Default, Clone, Copy, PartialEq, Eq, Debug)]
-pub struct Bitset(u8);
+pub struct Missing(u8);
 
-impl Bitset {
+impl Missing {
+    /// Returns `true` if the set contains the component index.
     pub fn contains(self, ix: usize) -> bool {
         (self.0 & (1 << ix)) != 0
     }
 
-    pub fn set(&mut self, ix: usize) {
+    /// Adds a component index to the set.
+    pub fn insert(&mut self, ix: usize) {
         self.0 |= 1 << ix;
     }
 
-    pub fn single(ix: usize) -> Self {
+    /// The set containing a single component index.
+    pub fn singleton(ix: usize) -> Self {
         Self(1 << ix)
     }
 
-    pub fn any(self) -> bool {
-        self.0 != 0
+    /// Returns `true` if the set contains no indices.
+    pub fn is_empty(self) -> bool {
+        self.0 == 0
     }
 }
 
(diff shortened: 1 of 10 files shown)


waywardmonkeys (approved):
Apart from the rebase error, it seems fine?

A lot of it is just boilerplate change, so hopefully I didn't miss anything in there.

color/src/missing.rs
```diff
@@ -3,45 +3,49 @@
...
 
-    pub fn set(&mut self, ix: usize) {
+    /// Adds a component index to the set.
+    pub fn insert(&mut self, ix: usize) {
         self.0 |= 1 << ix;
     }
 
-    pub fn single(ix: usize) -> Self {
+    /// The set containing a single component index.
+    pub fn singleton(ix: usize) -> Self {
```
nicoburns, line 22: `singleton` seems like a confusing name for this. This usually implies what in Rust would be a static instance. I would prefer the old name of `single`. Or perhaps an idiomatic rusty name would be `with_single_component`.

color/src/missing.rs
```diff
@@ -3,45 +3,49 @@
 
 //! A simple bitset.
 
-/// A simple bitset, for representing missing components.
+/// A simple bitset for representing missing components.
 #[derive(Default, Clone, Copy, PartialEq, Eq, Debug)]
-pub struct Bitset(u8);
+pub struct Missing(u8);
```
nicoburns, line 8: Would it be clearer if this represented present (non-missing) components? 0 to represent "present" and 1 to represent "missing" seems somehow inverted. Presumably this is equivalent on a technical level?

nicoburns: Perhaps "missing" semantics make sense because the common case is "no missing components"? If so then perhaps the code could stay the same, but the method could be documented in terms of "which components are present (non-missing)" rather than (or in addition to) "which components are in the set".

e.g. for `is_empty`:

```
Returns `true` if the set contains no indices (and therefore all color components are present).
```

nicoburns: Would also appreciate an/some example(s) of when components can be missing.

nicoburns (approved):
Broadly looks good to me. Some comments on the `Missing` type.

color/src/missing.rs
raphlinus: I've gone back to `single`, at your suggestion.

color/src/missing.rs
raphlinus: I'll add some docs about missing components.

Inverting the boolean sense is equivalent at a technical level. I think the strongest case for this way ("missing") is that the default value of 0 makes sense.

Incidentally, I considered and rejected some other representations for this. In particular, [color.js](https://colorjs.io/docs/interpolation) uses `NaN`. That works in pure Rust, but I have at least two problems with it: it complicates GPU interop, because NaN is almost always a bad thing), and it makes the test for "any missing components" more expensive, because you have to do a finiteness check on all the components, as opposed to comparison with 0 here. The current approach also has no memory cost, `DynamicColor` is 20 bytes either way.

PR #27, opened 2024-11-10
Add all docstrings

raphlinus:
Provide docstrings for all public items, and remove the exception. Going forward.

A couple of minor cleanups along the way.


waywardmonkeys (approved):
I think some places need more explanation, like hue fix up, but this is good for now!

PR #28, opened 2024-11-10
Use error enum.

### LINKED ISSUE #16: Real error types
`Error` from the parsing code should be a real error type rather than just a `&static str`.

waywardmonkeys:
This renames `Error` to `ParseError` and makes it an enumeration rather than a `&static str`.

This doesn't yet improve the data tracked for the error for better reporting.

Fixes #16.

### DIFF
--- a/color/src/parse.rs
+++ b/color/src/parse.rs
@@ -3,19 +3,95 @@
 
 //! Parse CSS4 color
 
+use core::error::Error;
 use core::f64;
+use core::fmt;
 use core::str::FromStr;
 
 use crate::{AlphaColor, ColorSpaceTag, DynamicColor, Missing, Srgb};
 
-// TODO: proper error type, maybe include string offset
+// TODO: maybe include string offset
 /// Error type for parse errors.
 ///
-/// Currently just a static string, but likely will be changed to
-/// an enum.
-///
 /// Discussion question: should it also contain a string offset?
-pub type Error = &'static str;
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum ParseError {
+    /// Unclosed comment
+    UnclosedComment,
+    /// Unknown angle dimension
+    UnknownAngleDimension,
+    /// Unknown angle
+    UnknownAngle,
+    /// Unknown color component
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This is meant to maintain the status quo in actual parsing error detection and reporting, just switching it over to using an enum.

We can file further issues for other improvements in the reporting, tracking more context, etc.

waywardmonkeys:
I'm inclined to move `ParseError` into a new file, perhaps at `color/src/parse/error.rs` but should the existing parse code stay in `color/src/parse.rs`?

I really don't like this code being at the top of this file ... nor at the end.

color/src/parse.rs
```diff
@@ -3,19 +3,95 @@
...
+    InvalidHexDigit,
+    /// Wrong number of hex digits
+    WrongNumberOfHexDigits,
+}
+
+impl Error for ParseError {}
+
+impl fmt::Display for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
```
raphlinus, line 52: You can make this a lot more compact by `let msg = match *self { ... }` then `f.write_str(msg)`.

color/src/parse.rs
```diff
@@ -3,19 +3,95 @@
...
+    ExpectedComma,
+    /// Invalid hex digit
+    InvalidHexDigit,
+    /// Wrong number of hex digits
+    WrongNumberOfHexDigits,
+}
+
+impl Error for ParseError {}
+
+impl fmt::Display for ParseError {
```
raphlinus, line 50: In general, I like to see the types first, then the impls, but not sure about that here. If you want to move it into a separate file, it'd be fine (and wouldn't require another approval).

raphlinus (approved):
Looks good, thanks!

I think we can consider this fine for 0.1, but there's a fair amount more to think about for parsing and error handling, including how to best fit into host syntax for example in `simplecess`.

color/src/parse.rs
raphlinus: I pushed this to the branch with Bruce's agreement.

PR #30, opened 2024-11-10
Hwb color space

### LINKED ISSUE #19: Add CSS4 color space HWB
See https://www.w3.org/TR/css-color-4/#the-hwb-notation

raphlinus:
Add HWB color space, basically following the spec. It's been spot-tested with non-HDR colors.

Serialization is to syntax which can be parsed by CSS Color 4, but doesn't include `%` for W and B, as is done by color.js. It also doesn't strictly follow serialization by the spec, which calls for the color to be lowered to RGB. To me this is an argument in favor of having a separate code path for strict CSS serialization.

In theory, HWB should be able to handle HDR colors with negative values for W and B. In practice, that's a bad idea, and in any case the current code does not roundtrip these correctly. That might be a flaw in the spec, and deserves deeper investigation.

Closes #19

### DIFF
--- a/color/examples/parse.rs
+++ b/color/examples/parse.rs
@@ -11,7 +11,7 @@
 //! cargo run --example parse 'oklab(0.5 0.2 0)'
 //! ```
 
-use color::{AlphaColor, Lab, Srgb};
+use color::{AlphaColor, Hwb, Lab, Srgb};
 
 fn main() {
     let arg = std::env::args().nth(1).expect("give color as arg");
@@ -23,6 +23,8 @@ fn main() {
             println!("{srgba:?}");
             let lab: AlphaColor<Lab> = color.to_alpha_color();
             println!("{lab:?}");
+            let hwb: AlphaColor<Hwb> = color.to_alpha_color();
+            println!("{hwb:?}");
         }
         Err(e) => println!("error: {e}"),
     }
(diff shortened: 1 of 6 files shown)


raphlinus:
A bit more on the HDR color issue. Using [color.js](https://github.com/color-js/color.js), this code:

```js
let c4 = new Color("color(srgb 1.1 1.2 1.1)");
console.log(c4.to("srgb").toString({inGamut: false}));
console.log(c4.to("hsl").toString({inGamut: false}));
console.log(c4.to("hwb").toString({inGamut: false}));
```

produces this output:

```
rgb(110% 120% 110%)
hsl(300 33.333% 115%)
hwb(120 110% -20%)
```

The hwb value is what I'd expect (120 is green), and this converts back correctly. The hsl value has a very strange hue, but does convert back correctly. I think the problem is that the [sample code in the spec](https://www.w3.org/TR/css-color-4/#rgb-to-hwb) uses the hue from HSL, and that's not valid when the hue rotation hack. That was noticed in https://github.com/w3c/csswg-drafts/issues/10695 and apparently fixed by [`ead8668`](https://github.com/w3c/csswg-drafts/commit/ead8668d310a164a8beef5af615e061a3cee95be), but I don't see the update in the spec.

waywardmonkeys (approved):

PR #35, opened 2024-11-12
Add Adobe RGB (1998) color space

tomcur:
The Linear A98 RGB <-> Linear sRGB matrices are based on the exact rational conversion matrices calculated from the XYZ chromaticities. This crate doesn't need the exact rationals as they're truncated to f32 at compile-time anyway, but it might be worth having to serve as a reference for other people. It would be nice to have agreement across implementations at any given precision.

The same _can_ be done for other basis change matrices, though it becomes somewhat unwieldy as the number of steps combined into the matrix increases; by my calculation element (0,0) of the Linear sRGB to XYZ-D50 matrix with linear Bradford adaptation is exactly `7206783369626736730083375524119595656 / 16526827482518743502029642682152162875`.

This is all based on the assumption that the color spaces' specified xy chromaticities are normative, and the specified matrices in the specs are computed off those.

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -283,6 +283,79 @@ impl ColorSpace for DisplayP3 {
     }
 }
 
+/// 🌌 The Adobe RGB (1998) color space.
+///
+/// Adobe RGB is similar to [sRGB](`Srgb`) but has higher green chromaticity, thereby extending its
+/// gamut over sRGB on that component. It was developed to encompass typical color print gamuts.
+///
+/// Its components are `[r, g, b]` (red, green, and blue channels respectively), with `[0, 0, 0]`
+/// pure black and `[1, 1, 1]` white. The natural bounds of the channels are `[0, 1]`.
+///
+/// This corresponds to the color space in [CSS Color Module Level 4 § 10.5][css-sec] and is
+/// [characterized by the ICC][icc]. Adobe RGB is described [here][adobe] by Adobe.
+///
+/// [css-sec]: https://www.w3.org/TR/css-color-4/#predefined-a98-rgb
+/// [icc]: https://www.color.org/chardata/rgb/adobergb.xalter
+/// [adobe]: https://www.adobe.com/digitalimag/adobergb.html
+#[derive(Clone, Copy, Debug)]
+pub struct A98Rgb;
+
+impl ColorSpace for A98Rgb {
+    const TAG: Option<ColorSpaceTag> = Some(ColorSpaceTag::A98Rgb);
+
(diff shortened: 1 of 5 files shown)


raphlinus:
The exact rationals work well here, but overall I think they're gilding the lily. Doing the calculations in f64 and then truncating to f32 as the absolute last step is what I've generally been doing, and I think is fine.

color/src/colorspace.rs
```diff
@@ -283,6 +283,79 @@ impl ColorSpace for DisplayP3 {
...
+            [0., 1., 0.],
+            [
+                0.,
+                (-11_512_411. / 268_173_353.) as f32,
+                (279_685_764. / 268_173_353.) as f32,
+            ],
+        ];
+        matmul(
+            &LINEAR_A98RGB_TO_SRGB,
+            [r, g, b].map(|x| x.abs().powf(2.199_218_8).copysign(x)),
```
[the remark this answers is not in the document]
tomcur: Yes, I think I accidentally cargo clippy'd it from the exact 2.19921875.

color/src/colorspace.rs
```diff
@@ -283,6 +283,78 @@ impl ColorSpace for DisplayP3 {
...
+            [0., 1., 0.],
+            [
+                0.,
+                (-11_512_411. / 268_173_353.) as f32,
+                (279_685_764. / 268_173_353.) as f32,
+            ],
+        ];
+        matmul(
+            &LINEAR_A98RGB_TO_SRGB,
+            [r, g, b].map(|x| x.abs().powf(563. / 256.).copysign(x)),
```
DJMcNab, line 327: Should this be `const { (563. / 256.) as f32 }`, similarly to the values above. That's a lot noisier than the current form...

tomcur: The above form requires the explicit cast to f32 so the compiler does the calculation (inside the parentheses) in f64, truncating to f32 at the last step. The reason for that is that the numerator and denominator have more significant digits than fit in a f32, so doing the entire thing in f32 may cause a very small, but avoidable, error.

For example, 

```
(-19_070_177. / 47_872_228.) as f32 -> #becbf549
-19_070_177f32 / 47_872_228f32      -> #becbf548
```

DJMcNab: Right, makes sense. I was just pattern matching from Raph's approval comment).

tomcur:
> I'm also slightly curious, what tool are you using? For my matrix multiplications I happen to be using the reference code from the CSS spec, but pretty much anything should work and will give the same results.

I've been using these conversions, and some work on Kurbo, as an excuse to learn [Sage](https://www.sagemath.org/). It's nice for these conversions as you can specify them algebraically, and if you plug in rationals, the calculations are done over the rationals.

Issue #26, opened 2024-11-09
Serialization of color names


raphlinus:
There's a basic question that needs a decision (though it doesn't have to happen right now), and some more subtle considerations.

The [serialization section](https://www.w3.org/TR/css-color-4/#serializing-color-values) of the CSS Color 4 spec requires that a parse of `"red"` serializes as `"red"`, not `"rgb(255, 0, 0)"` or `"#ff0000"`. I can see this as potentially useful in applications that process CSS for output to the web, improving human readability. We could do it, but it has some cost and complexity.

The outline of how I'd do it is reserve another bit in `Missing` (which I'd probably rename "Flags" or somesuch) indicating that it's a named color. Then, when serializing, if the bit is set, the color would be rounded to rgba8, the named color table would be searched, and, if there's a match, the name output. If there's not a match, then it would be serialized as if the bit were not set. The search could be implemented at least 3 ways: linear scan (slow-ish but simple and with no additional data burden), phf (tricky, estimated 284 bytes), or binary search using an additional 142 byte permutation table representing the sort by u32.

But this touches a deeper problem. There is some potentially some divergence between what we want for string output, and strict CSS serialization. Aside from the named colors, we need to serialize out-of-gamut srgb using the `color` function and `srgb` colorspace, rather than the `rgb` function, as the latter clips. Also, if we have color spaces not covered by CSS (#14, #15, possibly others), then those won't serialize to a valid CSS string.

One solution is that we do our own thing for `Display` on our color types that is not necessarily CSS compatible, and have another way to do strictly CSS compatible serialization. I'd lean toward doing that as `Display` impl on a `CssFormat` wrapper type, and also have that feature gated, but am open to other ideas.

DJMcNab:
Do we not have enough space available to just store which named colour it was inline? From a quick check, we seem to have 16 bits of padding in `DynamicColor` without even thinking of packing things in more tightly, which would allow 65000 named colours. Obviously you have to be careful about when to discard those, but you have to discard the named bit in the same circumstances anyway.

tomcur:
> Aside from the named colors, we need to serialize out-of-gamut srgb using the color function and srgb colorspace, rather than the rgb function, as the latter clips.

[...]

PR #39, opened 2024-11-13
Serialize color names by packing them into `DynamicColor`

tomcur:
To perform correct CSS Color 4 serialization of colors, some details of how a color was declared have to be preserved. Specifically, [CSS Color Module Level 4 § 15.2](https://www.w3.org/TR/css-color-4/#serializing-sRGB-values) and the two sections on Lab and Oklab following it specify that 
- a color declared as a named CSS/X11 color, should serialize to that (normalized) color name; and 
- a color declared as one of the named color space functions – i.e., _not_ the `color()` function – should serialize to that function. (Exceptions being `hsl()` and `hwb()`, those serialize to `rgb()`.)

This implementation uses the padding currently available on `DynamicColor` to pack in the color name the `DynamicColor` was created from, or whether a named color space function was used; an implementation of the ideas in https://github.com/linebender/color/issues/26#issuecomment-2467997421.

`DynamicColor` now contains `Flags`, and the existing `Missing` color components are moved into `Flags`. In this implementation, only `parse_color` sets the color name or color space function name.


color/src/serialize.rs
```diff
@@ -77,19 +77,61 @@ fn write_legacy_function(
...
+    fn specified_to_serialized() {
+        for (specified, expected) in [
+            ("rgb(1,1,1)", "rgb(1, 1, 1)"),
+            // currently fails, but should succeed (values should be clamped at parse-time)
+            // ("rgb(1.1,1,1)", "rgb(1, 1, 1)"),
+            ("color(srgb 1.0 1.0 1.0)", "color(srgb 1 1 1)"),
+            ("rosybrown", "rosybrown"),
+            ("red", "red"),
+            ("transparent", "transparent"),
+            ("yellowgreen", "yellowgreen"),
```
tomcur, line 127: I _think_ this is the correct behavior based on https://www.w3.org/TR/css-color-4/#serializing-color-values.

color/src/serialize.rs
```diff
@@ -77,19 +77,61 @@ fn write_legacy_function(
...
+        ] {
```
tomcur, line 128: I _think_ this is the correct behavior based on https://www.w3.org/TR/css-color-4/#serializing-color-values.

color/src/serialize.rs
```diff
@@ -77,19 +77,69 @@ fn write_legacy_function(
...
+            let result = format!("{}", parse_color(specified).unwrap());
+            assert_eq!(
+                result,
+                expected,
+                "Failed serializing specified color `{specified}`. Expected: `{expected}`. Got: `{result}`."
+            );
```
tomcur, line 142: I _think_ this is the correct behavior based on https://www.w3.org/TR/css-color-4/#serializing-color-values.

color/src/dynamic.rs
```diff
@@ -39,6 +41,11 @@ pub struct DynamicColor {
     pub components: [f32; 4],
 }
 
+// `DynamicColor` was carefully packed. Ensure its size doesn't accidentally change.
+const _: () = if size_of::<DynamicColor>() != 20 {
```
DJMcNab, line 45: This should probably be `cfg(test)`

color/src/serialize.rs
```diff
@@ -77,19 +77,69 @@ fn write_legacy_function(
...
+                "hwb(740deg 20% 30% / 50%)",
+                "rgba(178.5, 93.50008, 50.999996, 0.5)",
+            ),
+            // currently fails, but should succeed (values should be clamped at parse-time)
+            // ("rgb(1.1,1,1)", "rgb(1, 1, 1)"),
+            ("color(srgb 1.0 1.0 1.0)", "color(srgb 1 1 1)"),
+            ("rosybrown", "rosybrown"),
+            ("red", "red"),
+            ("transparent", "transparent"),
+            ("yellowgreen", "yellowgreen"),
```
DJMcNab, line 135: There's a simple exhaustive test that the named colours all roundtrip. I think it could only go wrong if the phf were wrong (?), but that would still be good to catch. I don't think that test would take *too* long to run.

DJMcNab (commented):
Looks good! Some more passing thoughts.

color/src/missing.rs
```diff
@@ -3,27 +3,124 @@
...
+/// The latter is primarily useful for serializing.
+///
+/// [css-named-colors]: https://www.w3.org/TR/css-color-4/#named-colors
+/// [css-named-color-spaces]: https://www.w3.org/TR/css-color-4/#color-syntax
+//
+// The flags are tracked with 16 bits. The first three are for missing components, the fourth
+// indicates whether the color was generated from a named color or named color space function. The
+// remaining bits indicate the named color.
+#[derive(Default, Clone, Copy, PartialEq, Eq, Debug)]
+pub struct Flags(u16);
```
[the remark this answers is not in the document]
waywardmonkeys: Raph had mentioned (somewhere) that he would rename `Missing` to `Flags` or something if he went down this path, I think.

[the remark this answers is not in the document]
tomcur: `DynamicColor` can fit 24 additional bits (of which `Missing` currently takes 8). That can be done with an 8-bit type and a 16-bit type. If ergonomics of constructing `DynamicColor`  doesn't matter too much, I think I would also prefer using two fields. Alternatively, Flags could become

```rust 
struct Flags {
    missing: u8,
    name: u8,
}
```

where, for example, `named == 0` indicates the value is not from a named color or named color space, `named == 255` indicates it is from a named color space, and other values are the 1-based index into `crate::x11_colors::NAMES`.

color/src/missing.rs
```diff
@@ -10,21 +10,26 @@ use crate::x11_colors;
...
 /// [css-named-color-spaces]: https://www.w3.org/TR/css-color-4/#color-syntax
-//
-// The flags are tracked with 16 bits. The first three are for missing components, the fourth
-// indicates whether the color was generated from a named color or named color space function. The
-// remaining bits indicate the named color.
 #[derive(Default, Clone, Copy, PartialEq, Eq, Debug)]
-pub struct Flags(u16);
+pub struct Flags {
+    /// A bitset of missing color components.
+    missing: u8,
```
DJMcNab, line 25: Would it make sense for this to be `Missing` directly?

DJMcNab (commented):
What's needed to move this out of draft?

color/src/missing.rs
```diff
@@ -3,27 +3,124 @@
...
+    /// Returns `true` if the flags indicate the color was generated from a named color or named
+    /// color space function.
+    pub fn named(self) -> bool {
+        self.0 & 1 << 4 != 0
+    }
+
+    /// If the color was constructed from a named color, returns that name.
+    ///
+    /// See also [`parse_color`][crate::parse_color].
+    pub fn color_name(self) -> Option<&'static str> {
```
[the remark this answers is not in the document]
tomcur: As Flags and Missing are opaque, we could `Debug` format to, for example:

```
Flags { data: 0b0000010000000000, missing: [false, true, false, false], named: true, color_name: Some("rosybrown") }
Missing { data: 0b00000010, missing: [false, true, false, false] }
```

color/src/missing.rs
tomcur: Yes, I think so.

To shrink the API surface, maybe also remove the `missing` methods from `Flags`? Except for

```rust
pub const fn from_missing(missing: Missing) -> Self
pub fn set_missing(&mut self, missing: Missing)
pub fn missing(self) -> Missing
```

Code would then call as e.g. `dynamic_image.flags.missing().contains(1)`, and using the `Missing` bitops would fit more naturally into that pattern.

What are your thoughts?

tomcur:
> What's needed to move this out of draft?

I think when https://github.com/linebender/color/pull/39#discussion_r1844964057 is resolved, we can move this out of draft.

color/src/missing.rs
DJMcNab: Seems reasonable enough.

color/src/serialize.rs
```diff
@@ -136,3 +154,47 @@ impl core::fmt::UpperHex for Rgba8 {
...
+                "rgba(178.5, 93.50008, 50.999996, 0.5)",
+            ),
+            // the next two currently fail, but should succeed (ASCII uppercase codepoints should
+            // be lowercased)
+            // ("ReD", "red"),
+            // ("RgB(1,1,1)", "rgb(1, 1, 1)"),
+            // currently fails, but should succeed (values should be clamped at parse-time)
+            // ("rgb(1.1,1,1)", "rgb(1, 1, 1)"),
+            ("color(srgb 1.0 1.0 1.0)", "color(srgb 1 1 1)"),
+        ] {
```
tomcur, line 179: I believe this is the correct serialization as specified by CSS Color 4. To leave case normalization out of the crate, callers wanting case-insensitivity could be instructed to [`str::make_ascii_lowercase`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.make_ascii_lowercase). Otherwise the parser requires some work to get performant insensitive matching.

raphlinus:
I'm having trouble making a go/no-go decision on this. I like the idea of better CSS compatibility, and the implementation seems fine (though I didn't dig deep), but I also like the idea of the code being minimal, and there are applications that will not need this functionality. I know we talked about this in office hours but don't feel consensus emerged.

Does anyone have a strong feeling either way? My gut feeling is that it's easier to add later if there's a strong need than remove, but if it is going in anyway, now is a good time to do it.

waywardmonkeys:
@tomcur They wouldn't necessarily have to do a wrapping type ...they could just do a very quick linear scan of the colors for a matching one. Sure, it is moderately gross, but it is something they'd opt into.

tomcur:
Not exactly, unfortunately. The CSS spec says the color parsed from the string `red` serializes to `red`, `rgb(1, 0, 0)` serializes to `rgb(1, 0, 0)` and `color(srgb 1 0 0)` serializes to `color(srgb 1 0 0)`. To be able to do that, some information about the source has to be retained, either in `DynamicColor` or as part of the return value from `parse_color`.

waywardmonkeys:
Good points. I lean towards this landing for 0.2. @raphlinus ?

raphlinus (approved):
I am ok with landing this, it seems like a nice convenience for users, and I'm not too worried about the cost.

nicoburns:
> I also like the idea of the code being minimal, and there are applications that will not need this functionality.

> The main argument to include it would be that it's not straightforward to implement on top of the library.

This sounds very much like it ought to exist behind a feature flag to me! (but I think the short term it doesn't matter too much - we can always and then feature flag it in the future version).

color/src/x11_colors.rs
```diff
@@ -312,15 +312,15 @@ fn weak_hash(key: u32, salt: u32, n: usize) -> usize {
     (((y as u64) * (n as u64)) >> 32) as usize
 }
 
-pub(crate) fn lookup_palette(s: &str) -> Option<[u8; 4]> {
+pub(crate) fn lookup_palette(s: &str) -> Option<usize> {
```
waywardmonkeys, line 315: Could you rename this to `lookup_palette_index` or similar? Maybe even with css_palette? Or should this not be duplicated and moved to the palette code with a single function definition?

tomcur: The rename: e227926c3ee048ea8afd8698b34ece04bd3a146e, 013c90b7c2b92d41c8b673d81bd2321da803c9fe.

Moving it around is a bit more difficult as the file is auto-generated, but I'm unsure what you mean by duplication of the function.

color/src/flags.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub const fn from_missing(missing: Missing) -> Self {
+        Self { missing, name: 0 }
+    }
+
+    /// Set the missing components.
+    #[inline]
+    #[warn(
+        clippy::missing_const_for_fn,
+        reason = "can be made const with MSRV 1.83"
+    )]
```
tomcur, line 60: I've added warning lints in 70c198396d4fcb5e1027cd97626b4060033d5cd4 as suggested in #91, but interestingly Clippy 0.1.83 doesn't warn. I did bump `workspace.package.rust-version` in `Cargo.toml`.

DJMcNab: Hmm. I guess it doesn't know about the mutability change.

tomcur:
Rebasing is done, but I also made some larger changes to `color/src/serialize.rs` especially. It's along the same lines, but fixes some edge cases. I think another look from someone before merging is a good idea.

color/src/flags.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    /// Returns the missing components from the flags.
+    #[inline]
+    pub const fn missing(self) -> Missing {
+        self.missing
+    }
+
+    /// Set the flags to indicate the color was specified as one of the named colors. `name_ix` is
+    /// the index into [`crate::x11_colors::NAMES`].
+    pub(crate) fn set_named_color(&mut self, name_ix: usize) {
```
DJMcNab, line 73: This will never be called from user code, right? (I ask because otherwise `color_name` could panic in release mode)

color/src/parse.rs
```diff
@@ -461,31 +461,53 @@ impl<'a> Parser<'a> {
...
     if let Some(id) = parser.ident() {
         let color = match id {
-            "rgb" | "rgba" => parser.rgb(),
-            "lab" => parser.lab(100.0, 1.25, ColorSpaceTag::Lab),
-            "lch" => parser.lch(100.0, 1.25, ColorSpaceTag::Lch),
-            "oklab" => parser.lab(1.0, 0.004, ColorSpaceTag::Oklab),
-            "oklch" => parser.lch(1.0, 0.004, ColorSpaceTag::Oklch),
-            "hsl" | "hsla" => parser.hsl(),
-            "hwb" => parser.hwb(),
+            "rgb" | "rgba" => parser.rgb().map(set_from_named_color_space),
```
DJMcNab, line 482: Not a huge fan of this pattern, but I'm not sure how to avoid it...

DJMcNab (approved):
I've not reasoned about this too carefully, but the code quality is excellent

color/src/flags.rs
tomcur: That's right, only the parser calls this.

color/src/parse.rs
tomcur: You mean the `map(set_from_named_color_space)`? I agree; I remember I tried some alternatives and that this pattern was the least bad.

PR #41, opened 2024-11-14
Implementation of Rgba8 type

### LINKED ISSUE #24: Rgba8
Packed u8 color type

raphlinus:
Adds Rgba8 type and some conversions and formatting.

This PR also changes the serialization of srgb to use the color function. If legacy RGB serialization is desired, the client can convert to Rgba8 first.

Closes #24.

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -6,7 +6,7 @@
 use core::any::TypeId;
 use core::marker::PhantomData;
 
-use crate::{ColorSpace, ColorSpaceLayout, ColorSpaceTag, Oklab, Oklch};
+use crate::{ColorSpace, ColorSpaceLayout, ColorSpaceTag, Oklab, Oklch, Rgba8, Srgb};
 
 #[cfg(all(not(feature = "std"), not(test)))]
 use crate::floatfuncs::FloatFuncs;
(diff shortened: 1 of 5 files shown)


raphlinus:
I'm pushing this partly for discussion. There's not a huge amount of functionality attached to the new type - maybe that's a good thing, we should encourage all actual manipulation (including parsing) to happen in with the f32-based types.

waywardmonkeys (approved):
I don't even see anything worth discussing offhand. Seems reasonable.

I guess the one thing might be if we wanted to have an `Rgb8` without alpha, but I don't think that's necessary and if it was, experiment with it in the codebase that needs it.

The one thing that I see that might be missing is a `to_premul_u32` method like `peniko::Color` has (and that Vello uses). But Vello can also just do that math on its own...

tomcur:
> This PR also changes the serialization of srgb to use the color function. If legacy RGB serialization is desired, the client can convert to Rgba8 first.

I believe that follows the spec, which does that _unless_ the color was constructed directly from a named color or from a named color space ([if I'm reading this correctly](https://www.w3.org/TR/css-color-4/#serializing-sRGB-values) – note it excludes the `color` function). The named color space functions that map to RGB clamp during parse-time according to the spec, so can be represented as legacy RGB when serialized.

i.e.,

`rgba(50, 75, 100, 0.8)` serializes to `rgba(50, 75, 100, 0.8)`
`color(srgb 0.1 0.2 0.3 / 0.8)` serializes to `color(srgb 0.1 0.2 0.3 / 0.8)`

In my experimental draft #39 I'm trying to find a nice implementation according to spec.

raphlinus:
Great, I'll merge this and then we can consider further refinements separately.

PR #42, opened 2024-11-14
Add Rec2020 color space

### LINKED ISSUE #20: CSS 4: Predefined color spaces
Will we support all of the CSS 4 predefined color spaces?

See https://www.w3.org/TR/css-color-4/#predefined

tomcur:
Resolves #20, now implementing all CSS Color Module Level 4 color spaces.

Further:

- Add Tom Churchman to AUTHORS list.

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -421,6 +421,110 @@ impl ColorSpace for ProphotoRgb {
     }
 }
 
+/// 🌌 The Rec. 2020 color space.
+///
+/// Rec. 2020 is similar to [sRGB](`Srgb`) but has higher red, green and blue chromaticities,
+/// thereby extending its gamut over sRGB on all components.
+///
+/// Its components are `[r, g, b]` (red, green, and blue channels respectively), with `[0, 0, 0]`
+/// pure black and `[1, 1, 1]` white. The natural bounds of the channels are `[0, 1]`.
+///
+/// This corresponds to the color space in [CSS Color Module Level 4 § 10.7][css-sec] and is
+/// [characterized by the ICC][icc]. The color space is defined by the International
+/// Telecommunication Union [here][itu].
+///
+/// [css-sec]: https://www.w3.org/TR/css-color-4/#predefined-rec2020
+/// [icc]: https://www.color.org/chardata/rgb/BT2020.xalter
+/// [itu]: https://www.itu.int/rec/R-REC-BT.2020/en
+#[derive(Clone, Copy, Debug)]
+pub struct Rec2020;
+
+impl Rec2020 {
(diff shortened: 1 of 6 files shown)


color/src/colorspace.rs
```diff
@@ -421,6 +421,110 @@ impl ColorSpace for ProphotoRgb {
...
+            ],
+            [
+                (-4_638_020_506. / 37_238_079_773.) as f32,
+                (42_187_016_744. / 37_238_079_773.) as f32,
+                (-310_916_465. / 37_238_079_773.) as f32,
+            ],
+            [
+                (-97_469_024. / 5_369_968_309.) as f32,
+                (-3_780_738_464. / 37_589_778_163.) as f32,
+                (42_052_799_795. / 37_589_778_163.) as f32,
```
tomcur, line 472: Some more lily-gilding here. Debatable usefulness, as the transfer function parameters are reals not representable as rationals.

color/src/colorspace.rs
```diff
@@ -421,6 +421,110 @@ impl ColorSpace for ProphotoRgb {
...
+            ],
+        ];
+
+        fn transfer(x: f32) -> f32 {
+            if x.abs() < Rec2020::B * 4.5 {
+                x / 4.5
```
raphlinus, line 478: Should be x * (1.0 / 4.5) for performance.

color/src/colorspace.rs
```diff
@@ -421,6 +421,110 @@ impl ColorSpace for ProphotoRgb {
...
+                (1_772_384_008. / 20_137_682_025.) as f32,
+                (18_035_212_433. / 20_137_682_025.) as f32,
+            ],
+        ];
+
+        fn transfer(x: f32) -> f32 {
+            if x.abs() < Rec2020::B {
+                x * 4.5
+            } else {
+                (Rec2020::A * x.abs().powf(0.45) - Rec2020::A + 1.).copysign(x)
```
raphlinus, line 517: If you group it so it's `(Rec2020::A - 1.)` that's one fewer addition.

raphlinus (approved):
A couple of picky micro-optimizations, otherwise looks good. Again, spot-checked the color pairs in the test against color.js.

color/src/colorspace.rs
tomcur: I like how consistently you look for these; I'm learning to remember.

raphlinus:
Btw, we can now delete the `allow` for todos. That could have been done in this PR, but I'll do it in an upcoming one that improves docstrings.

DJMcNab:
Fwiw, that's another key advantage for `expect` - the self-cleansing aspect.

PR #43, opened 2024-11-15
Improve interop with `bytemuck` and FFI.

waywardmonkeys:
`AlphaColor`, `OpaqueColor`, and `PremulColor` are now using `#[repr(transparent)]` and impl the `bytemuck` traits `Pod`, `TransparentWrapper`, and `Zeroable`.

`Rgba8` is now `#[repr(C)]` and impls the `bytemuck` traits `Pod` and `Zeroable`.

The `bytemuck` traits are only available when the new `bytemuck` feature is enabled. It is not enabled by default.

This requires relaxing the forbidding of `unsafe_code` as the impls of the `bytemuck` traits must use `unsafe impl`.

### DIFF
--- a/color/Cargo.toml
+++ b/color/Cargo.toml
@@ -22,8 +22,14 @@ targets = []
 default = ["std"]
 std = []
 libm = ["dep:libm"]
+bytemuck = ["dep:bytemuck"]
 
 [dependencies]
+
+[dependencies.bytemuck]
+version = "1.19.0"
+optional = true
+default-features = false
 
 [dependencies.libm]
 version = "0.2.11"
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -22,6 +22,7 @@ use crate::floatfuncs::FloatFuncs;
 /// for spline interpolation. For cylindrical color spaces, hue fixup should
 /// be applied before interpolation.
 #[derive(Clone, Copy, Debug)]
+#[repr(transparent)]
 pub struct OpaqueColor<CS> {
     /// The components, which may be manipulated directly.
     ///
@@ -37,6 +38,7 @@ pub struct OpaqueColor<CS> {
 ///
 /// See [`OpaqueColor`] for a discussion of arithmetic traits and interpolation.
 #[derive(Clone, Copy, Debug)]
+#[repr(transparent)]
 pub struct AlphaColor<CS> {
     /// The components, which may be manipulated directly.
     ///
@@ -58,6 +60,7 @@ pub struct AlphaColor<CS> {
 ///
 /// See [`OpaqueColor`] for a discussion of arithmetic traits and interpolation.
 #[derive(Clone, Copy, Debug)]
+#[repr(transparent)]
 pub struct PremulColor<CS> {
     /// The components, which may be manipulated directly.
     ///
--- a/color/src/impl_bytemuck.rs
+++ b/color/src/impl_bytemuck.rs
@@ -1,0 +1,131 @@
+// Copyright 2024 the Color Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![allow(unsafe_code, reason = "unsafe is required for bytemuck unsafe impls")]
+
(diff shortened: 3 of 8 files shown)


color/Cargo.toml
```diff
@@ -22,9 +22,15 @@ targets = []
...
 std = []
 libm = ["dep:libm"]
+bytemuck = ["dep:bytemuck"]
 
 [dependencies]
 
+[dependencies.bytemuck]
+version = "1.19.0"
+optional = true
+default-features = false
```
DJMcNab, line 32: I'm sure there's a fallacy which applies to this argument, but nevertheless I'm going to make it. If our users want to use bytemuck, how likely is it that they also won't have the derive feature enabled?

That is, do we gain much value from being so austere?

color/src/color.rs
```diff
@@ -22,6 +22,7 @@ use crate::floatfuncs::FloatFuncs;
 /// for spline interpolation. For cylindrical color spaces, hue fixup should
 /// be applied before interpolation.
 #[derive(Clone, Copy, Debug)]
+#[repr(C)]
```
DJMcNab, line 25: I think `repr(transparent)` would be more accurate here; it would ensure that a new field being added requires more careful thought.

color/src/impl_bytemuck.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2024 the Color Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![allow(unsafe_code, reason = "unsafe is required for bytemuck unsafe impls")]
+
+use crate::{AlphaColor, ColorSpace, OpaqueColor, PremulColor, Rgba8};
+
+unsafe impl<CS: ColorSpace> bytemuck::Pod for AlphaColor<CS> {}
```
DJMcNab, line 8: Missing safety comment. That comment is trivial one we have the transparent repr.

color/src/impl_bytemuck.rs
```diff
@@ -0,0 +1,22 @@
...
+
+unsafe impl<CS: ColorSpace> bytemuck::Zeroable for OpaqueColor<CS> {}
+
+unsafe impl<CS: ColorSpace> bytemuck::Pod for PremulColor<CS> {}
+
+unsafe impl<CS: ColorSpace> bytemuck::Zeroable for PremulColor<CS> {}
+
+unsafe impl bytemuck::Pod for Rgba8 {}
+
+unsafe impl bytemuck::Zeroable for Rgba8 {}
```
DJMcNab, line 22: Can we have a "test" that all fields of these types (i.e. by doing destructuring) are either `std::marker::PhantomData` or are themselves Pod?

color/src/color.rs
raphlinus: Yes, I think `repr(transparent)` is right here. I wasn't sure `[f32; 4]` would be valid to cross the C ABI without the explicit `repr(C)`, but I think it is.

raphlinus (approved):
This seems like a good way to go. It's a bit sad that we have to trade off verifiable safety with keeping dependencies light, but I think it's the right decision. And hopefully when safe transmute happens we can go back to being safe.

I think the other things is that the crate docstring (and hence the README, thanks to cargo-rdme) talks about the features, so bytemuck should probably be added there.

color/src/impl_bytemuck.rs
waywardmonkeys: I've done something here.

raphlinus:
Do you also need to implement the [`TransparentWrapper`](https://docs.rs/bytemuck/latest/bytemuck/trait.TransparentWrapper.html) trait to be strictly conforming with bytemuck's expecations?

waywardmonkeys:
> Do you also need to implement the [`TransparentWrapper`](https://docs.rs/bytemuck/latest/bytemuck/trait.TransparentWrapper.html) trait to be strictly conforming with bytemuck's expecations?

I did when I switched to `repr(transparent)`.

PR #45, opened 2024-11-16
A first cut at module docstring

raphlinus:
Discuss some of the goals of the crate.

### DIFF
--- a/color/README.md
+++ b/color/README.md
@@ -26,6 +26,46 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
 
 Color is a Rust crate which implements color space conversions, targeting at least CSS4 color.
 
+## Scope and goals
+
+Color in its entirety is an extremely deep and complex topic. It is completely impractical
+for a single crate to meet all color needs. The goal of this one is to strike a balance,
+providing color capabilities while also keeping things simple and efficient.
+
+The main purpose of this crate is to provide a good set of types for representing colors,
+along with conversions between them and basic manipulations, especially interpolation. A
+major inspiration is the CSS Color Level 4 draft spec; we implement most of the operations
+and strive for correctness.
+
+Simplifications include:
+  * Always using `f32` to represent component values.
+  * Only handling 3-component color spaces (plus optional alpha).
+  * Choosing a fixed, curated set of color spaces for dynamic color types.
+  * Choosing linear sRGB as the central color space.
+  * Keeping white point implicit.
+
+A number of other tasks are out of scope for this crate:
+  * Print color spaces (CMYK).
+  * Spectral colors.
+  * Color spaces with more than 3 components generally.
+  * [ICC] color profiles.
+  * [ACES] color transforms.
+  * Appearance models and other color science not needed for rendering.
+  * Quantizing and packing to lower bit depths.
+
+## Main types
+
+The crate has two approaches to representing color in the Rust type system: a set of
+types with static color space as part of the types, and [`DynamicColor`] in which the
+color space is represented at runtime.
+
+The static color types come in three variants: [`OpaqueColor`] without an alpha channel,
+[`AlphaColor`] with a separate alpha channel, and [`PremulColor`] with premultiplied
+alpha. The last type is particularly useful for making interpolation and compositing more
+efficient. All have a generic type parameter with a trait bound of [`ColorSpace`], a
+zero sized type. The static types are open-ended, as it's possible to implement this
+trait for new color spaces.
+
 ## Features
 
 - `std` (enabled by default): Get floating point functions from the standard library (likely using your target's libc).
@@ -33,6 +73,9 @@ Color is a Rust crate which implements color space conversions, targeting at lea
 - `bytemuck`: Implement traits from `bytemuck` on [`AlphaColor`], [`OpaqueColor`], [`PremulColor`], and [`Rgba8`].
 
 At least one of `std` and `libm` is required; `std` overrides `libm`.
+
+[ICC]: https://color.org/
+[ACES]: https://acescentral.com/
 
 <!-- cargo-rdme end -->
 
--- a/color/README.md
+++ b/color/README.md
@@ -27,7 +27,7 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
 [`PremulColor`]: https://docs.rs/color/latest/color/struct.PremulColor.html
 [`OpaqueColor`]: https://docs.rs/color/latest/color/struct.OpaqueColor.html
 [`Rgba8`]: https://docs.rs/color/latest/color/struct.Rgba8.html
-[`ColorSpace`]: https://docs.rs/color/latest/color//trait.ColorSpace.html
+[`ColorSpace`]: https://docs.rs/color/latest/color/trait.ColorSpace.html
 <!-- cargo-rdme start -->
 
 Color is a Rust crate which implements color space conversions, targeting at least
--- a/color/README.md
+++ b/color/README.md
@@ -22,6 +22,12 @@ Full documentation at https://github.com/orium/cargo-rdme -->
 See https://linebender.org/blog/doc-include/ for related discussion. -->
 [libm]: https://crates.io/crates/libm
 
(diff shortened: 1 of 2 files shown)


color/README.md
```diff
@@ -26,6 +26,46 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
...
+  * Color spaces with more than 3 components generally.
+  * [ICC] color profiles.
+  * [ACES] color transforms.
+  * Appearance models and other color science not needed for rendering.
+  * Quantizing and packing to lower bit depths.
+
+## Main types
+
+The crate has two approaches to representing color in the Rust type system: a set of
+types with static color space as part of the types, and [`DynamicColor`] in which the
```
DJMcNab, line 59: These links should be evaluated and linked to in the block above (starting with `[libm]: ...`). You can use rust-analyzer to get the future docs.rs links for each item - see https://linebender.org/blog/doc-include/

color/README.md
```diff
@@ -26,6 +26,46 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
 
 Color is a Rust crate which implements color space conversions, targeting at least CSS4 color.
```
DJMcNab, line 27: This sentence was a stub added to help make the README. Maybe it's optimal as-is, but I would like affirmitive confirmation of that whilst we're here.

color/src/lib.rs
```diff
@@ -9,13 +9,49 @@
...
+//!
+//! The crate has two approaches to representing color in the Rust type system: a set of
+//! types with static color space as part of the types, and [`DynamicColor`] in which the
+//! color space is represented at runtime.
+//!
+//! The static color types come in three variants: [`OpaqueColor`] without an alpha channel,
+//! [`AlphaColor`] with a separate alpha channel, and [`PremulColor`] with premultiplied
+//! alpha. The last type is particularly useful for making interpolation and compositing more
+//! efficient. All have a generic type parameter with a trait bound of [`ColorSpace`], a
+//! zero sized type. The static types are open-ended, as it's possible to implement this
```
DJMcNab, line 52: Something about the sentence "All have a generic type parameter with a trait bound of `ColorSpace`, a zero sized type" doesn't quite sit right with me. I think the issue I'm having is that it explains concretely what is there, but not what this achieves.

Maybe something like, "These have a marker type parameter, indicating which [`ColorSpace`] they are in. Conversion to another colorspace uses the `convert` method on each of these types."

color/src/lib.rs
```diff
@@ -9,13 +9,49 @@
...
+//!   * Choosing linear sRGB as the central color space.
+//!   * Keeping white point implicit.
+//!
+//! A number of other tasks are out of scope for this crate:
+//!   * Print color spaces (CMYK).
+//!   * Spectral colors.
+//!   * Color spaces with more than 3 components generally.
+//!   * [ICC] color profiles.
+//!   * [ACES] color transforms.
+//!   * Appearance models and other color science not needed for rendering.
```
DJMcNab, line 39: I think this is a slightly buried lede; we should be upfront that our goal is computer graphics rendering.

color/src/lib.rs
```diff
@@ -9,13 +9,49 @@
...
+//!   * Quantizing and packing to lower bit depths.
```
DJMcNab, line 40: I think I know what you mean here (not dealing with f16?), but the Rgba type does exist, which I think does do this?

DJMcNab (commented):
Some drive-by comments. This looks really good!

color/README.md
raphlinus: I'm trying to follow those directions and having trouble. If I do this:

```
//! [AlphaColor]: crate::AlphaColor
```

Then I get ``warning: Could not resolve definition of `crate::AlphaColor`.`` If I leave that out, rdme doesn't complain, but I have no idea whether that would resolve in crates.io.

color/README.md
```diff
@@ -26,6 +26,46 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
...
+and strive for correctness.
+
+Simplifications include:
+  * Always using `f32` to represent component values.
+  * Only handling 3-component color spaces (plus optional alpha).
+  * Choosing a fixed, curated set of color spaces for dynamic color types.
+  * Choosing linear sRGB as the central color space.
+  * Keeping white point implicit.
+
+A number of other tasks are out of scope for this crate:
```
waywardmonkeys, line 47: I think it is worth saying that some of these may be added as other crates within the `color` ecosystem (of course, some can not be).

color/README.md
DJMcNab: Apologies. What I'm trying to indicate is quite subtle.
The section of the linked page I am referring to is the "Getting external documentation links" section.

To make these links resolve properly, we can have a "fallback" docs.rs link in the README, which will be used by any renderer of the GitHub README.

Using cargo-rdme means we are effectively doing the opposite of the trick described in the post.

color/README.md
tomcur: As a minimal change, I would write "CSS Color Module Level 4", "CSS Color Level 4", or "CSS Color 4" instead of "CSS4 color", to stick closer to the official name – and link that to `https://www.w3.org/TR/css-color-4/`.

color/src/lib.rs
```diff
@@ -9,13 +9,49 @@
...
+//! The crate has two approaches to representing color in the Rust type system: a set of
+//! types with static color space as part of the types, and [`DynamicColor`] in which the
+//! color space is represented at runtime.
+//!
+//! The static color types come in three variants: [`OpaqueColor`] without an alpha channel,
+//! [`AlphaColor`] with a separate alpha channel, and [`PremulColor`] with premultiplied
+//! alpha. The last type is particularly useful for making interpolation and compositing more
+//! efficient. All have a generic type parameter with a trait bound of [`ColorSpace`], a
+//! zero sized type. The static types are open-ended, as it's possible to implement this
+//! trait for new color spaces.
```
tomcur, line 53: Maybe another paragraph detailing `DynamicColor`. It is shortly touched on in the simplifications mentioned above. `ColorSpaceTag` enumerates the color spaces we support distinguishing at runtime, so arguably it is also a main type. Perhaps:

"In a `DynamicColor` the color space is represented at runtime using a `ColorSpaceTag`. Each color space implemented in this crate has a corresponding tag. Unlike the the static types, these tags are not open-ended."

color/src/lib.rs
```diff
@@ -9,13 +9,49 @@
...
-// these allow attributes.
-#![allow(clippy::todo, reason = "need to fix todos")]
 
 //! Color is a Rust crate which implements color space conversions, targeting at least CSS4 color.
 //!
+//! ## Scope and goals
+//!
+//! Color in its entirety is an extremely deep and complex topic. It is completely impractical
+//! for a single crate to meet all color needs. The goal of this one is to strike a balance,
+//! providing color capabilities while also keeping things simple and efficient.
```
tomcur, line 19: This is great.

tomcur (approved):
This looks good. A few small take-it-or-leave-it comments.

color/README.md
```diff
@@ -22,18 +22,83 @@ Full documentation at https://github.com/orium/cargo-rdme -->
 See https://linebender.org/blog/doc-include/ for related discussion. -->
 [libm]: https://crates.io/crates/libm
 
+[`DynamicColor`]: https://docs.rs/color/latest/color/struct.DynamicColor.html
+[`AlphaColor`]: https://docs.rs/color/latest/color/struct.AlphaColor.html
+[`PremulColor`]: https://docs.rs/color/latest/color/struct.PremulColor.html
+[`OpaqueColor`]: https://docs.rs/color/latest/color/struct.OpaqueColor.html
+[`Rgba8`]: https://docs.rs/color/latest/color/struct.Rgba8.html
+[`ColorSpace`]: https://docs.rs/color/latest/color//trait.ColorSpace.html
```
DJMcNab, line 30: Mentioned on Zulip, but for posterity:
```suggestion
[`ColorSpace`]: https://docs.rs/color/latest/color/trait.ColorSpace.html
```

DJMcNab (approved):

PR #50, opened 2024-11-17
Add `discard_alpha` utility function.

### LINKED ISSUE #46: Shorthand for discarding alpha
As I start to use the newly added `color::palette::css` code, I find that I'm often having to discard the alpha as I'm using `OpaqueColor` a lot.

I could do:

```
let (c, _) = color::palette::css::GOLDENROD.split();
```

But it would be nice for `AlphaColor` to have a const function for this like `discard_alpha` or `to_opaque`.

Either way, it isn't a big deal, just a small paper cut.

waywardmonkeys:
This is implemented for both `AlphaColor` and `PremulColor` to return an `OpaqueColor`.

In the case of a `PremulColor`, it will un-premultiply the alpha.

To keep this `const fn`, the `PremulColor::un_premultiply` function had to be marked as `const fn`.

Fixes #46.

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -307,6 +307,14 @@ impl<CS: ColorSpace> AlphaColor<CS> {
         (OpaqueColor::new(opaque), alpha)
     }
 
+    /// Split out the opaque components, discarding the alpha.
+    ///
+    /// This is a shorthand for calling [`split`](Self::split).
+    #[must_use]
+    pub const fn discard_alpha(self) -> OpaqueColor<CS> {
+        self.split().0
+    }
+
     /// Convert a color into a different color space.
     #[must_use]
     pub fn convert<TargetCs: ColorSpace>(self) -> AlphaColor<TargetCs> {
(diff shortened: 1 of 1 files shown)


color/src/color.rs
```diff
@@ -426,6 +434,15 @@ impl<CS: ColorSpace> PremulColor<CS> {
         Self { components, cs }
     }
 
+    /// Split out the opaque components, discarding the alpha.
+    ///
+    /// This is a shorthand for un-premultiplying the alpha and
+    /// calling [`AlphaColor::discard_alpha`].
```
raphlinus, line 440: I think I might mention that the result of this method on a fully transparent color is black.

color/src/color.rs
```diff
@@ -442,7 +459,7 @@ impl<CS: ColorSpace> PremulColor<CS> {
 
     /// Convert a color to the corresponding separate alpha form.
     #[must_use]
-    pub fn un_premultiply(self) -> AlphaColor<CS> {
+    pub const fn un_premultiply(self) -> AlphaColor<CS> {
```
raphlinus, line 462: :+1:

We might look for other opportunities to add `const`. In this case, it's a gimme, as `premultiply` is also const, so I would say that not having it was an oversight, corrected by this PR.

raphlinus (approved):
Looks good, thanks!

color/src/color.rs
waywardmonkeys: Added.

PR #54, opened 2024-11-22
Add ACEScg color space

### LINKED ISSUE #14: Add color space for ACEScg
See:

* https://docs.acescentral.com/specifications/acescg/

Also:

* https://acescolorspace.com
* https://www.chaos.com/blog/everything-you-need-to-know-about-acescg

MightyBurger:
This pull request resolves #14. See that issue topic and [this Zulip topic](https://xi.zulipchat.com/#narrow/channel/466849-color/topic/White.20Point.20and.20Color.20Transforms) for more information, especially on how that matrix was calculated.

I believe this is the first color space added to this crate that is not part of the CSS 4 spec. Because of this, I'm not sure how parsing or serialization should be done for this color space. Guidance would be appreciated, though we could possibly address this in a future PR.


color/src/colorspace.rs
```diff
@@ -525,6 +525,59 @@ impl ColorSpace for Rec2020 {
...
 }
 
+/// 🌌 The ACEScg color space.
+///
+/// This color space is defined by the Academy Color Encoding System specification [here][acescg].
+///
+/// The ACEScg color space is a linear color space. The wide gamut makes this color space useful as
+/// a working space for computer graphics.
+///
+/// Other ACES color spaces include ACES2065-1, ACEScc, and ACEScct.
```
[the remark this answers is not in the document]
MightyBurger: I added a note: "This crate currently does not support these color spaces."

color/src/colorspace.rs
```diff
@@ -525,14 +525,19 @@ impl ColorSpace for Rec2020 {
     }
 }
 
+#[expect(
+    clippy::doc_markdown,
+    reason = "these color space names should not be surrounded in backticks"
+)]
```
waywardmonkeys, line 531: Instead, you can add a `.clippy.toml` in the root of this repository with this line in it:

```
doc-valid-idents = ["ACEScc", "ACEScct", ".."]
```

Not sure if you'll need something for ACES2065-1.

color/src/colorspace.rs
```diff
@@ -525,6 +525,64 @@ impl ColorSpace for Rec2020 {
...
+#[derive(Clone, Copy, Debug)]
+pub struct AcesCg;
+
+impl ColorSpace for AcesCg {
+    const IS_LINEAR: bool = true;
+
+    const TAG: Option<ColorSpaceTag> = Some(ColorSpaceTag::AcesCg);
+
+    fn to_linear_srgb(src: [f32; 3]) -> [f32; 3] {
+        // XYZ_to_lin_sRGB * ACESwp_to_D65 * ACEScg_to_XYZ
```
waywardmonkeys, line 558: Should you use the exact rationals with the same `expect` thing that Rec2020 does?

MightyBurger: Tom used [Sage](https://www.sagemath.org/) to get those exact rational numbers. I don't know Sage, so it would take a fair bit more work for me to get those exact rationals. That work probably isn't worth it to "[guild the lily](https://github.com/linebender/color/pull/35#issuecomment-2474734716)" as Raph called it, since the resulting matrix should be the same.

I would guess the motivation for exact rationals is so someone could later calculate a more precise version of the matrix. As an alternative, providing documentation with enough information someone could derive the same result would serve the same purpose and would also be more informative.

color/src/colorspace.rs
```diff
@@ -525,6 +525,60 @@ impl ColorSpace for Rec2020 {
...
+        const ACESCG_TO_LINEAR_SRGB: [[f32; 3]; 3] = [
+            [1.705_051, -0.621_792_12, -0.083_258_87],
+            [-0.130_256_42, 1.140_804_8, -0.010_548_32],
+            [-0.024_003_36, -0.128_968_98, 1.152_972_3],
+        ];
```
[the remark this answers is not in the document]
MightyBurger: Yep, it was just a rounding error caused by rounding to eight decimal places in my calculations since I wasn't aware of Clippy's excessive precision lint. I re-ran the calculations and got the same result as you.

color/src/colorspace.rs
MightyBurger: Marking as resolved, let me know if you disagree. I can add more documentation, but there's more documentation in general I'd like to add, which can be done in a later PR.

waywardmonkeys:
This will need some rebasing and support for the new `WHITE_COMPONENTS` stuff.

color/src/colorspace.rs
```diff
@@ -525,6 +525,61 @@ impl ColorSpace for Rec2020 {
...
+/// a working space for computer graphics.
+///
+/// Other ACES color spaces include ACES2065-1, ACEScc, and ACEScct. This crate currently does not
+/// support these color spaces.
+///
+/// The ACEScg components are `[R, G, B]`. The components are bounded to `[-65504.0, 65504.0]`,
+/// though it is unusual to clip in this color space.
+///
+/// The conversion between D65 and the ACES white point is done using the standard Bradford
+/// linear chromatic adaptation transform. See the [XYZ-D65 color space](`XyzD65`) documentation for
```
tomcur, line 542: The choice so far has been to not go into too much detail about white point in the individual color spaces, Raph expressed that preference. We've used XYZ-D65 as a place to expand on the topic a bit. I think you can copy the wording from for example ProPhoto RGB about the white point.

```
/// ProPhoto RGB has a reference white of
/// D50; see the [XYZ-D65 color space](`XyzD65`) documentation for some background information on
/// the meaning of "reference white."
```

CHANGELOG.md
```diff
@@ -34,7 +35,9 @@ This release has an [MSRV][] of 1.82.
 This is the initial release.
 
 [@waywardmonkeys]: https://github.com/waywardmonkeys
+[@MightyBurger]: https://github.com/MightyBurger
```
waywardmonkeys, line 38: One tiny thing: alphabetical order here, so put you before me.

waywardmonkeys (approved):

tomcur (approved):
Looks good!

PR #55, opened 2024-11-22
Add `map_lightness` example

tomcur:


waywardmonkeys:
I liked your `swatch!` thing and wish it was here perhaps. Maybe.

DJMcNab (approved):
Seems useful. The same concern with the other examples that there are other versions of these methods. But not wanting to repeat the same example is also fine.

tomcur:
We could repeat the examples, or change #56 to link to them. I'll just merge this, and we can decide in #56.

tomcur:
> I don't think we want to block this on swatches, but I'd be happy to?

I think we haven't really made a decision yet whether it's nice enough to add. I also don't want to block this on that.

PR #56, opened 2024-11-22
Refer some AlphaColor method docs to OpaqueColor

tomcur:
This links to the examples, and prevents having to duplicate/keep-in-sync the documentation.

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -260,7 +260,7 @@ impl<CS: ColorSpace> OpaqueColor<CS> {
     /// let lighter = color.map_lightness(|l| l + 0.2);
     /// let expected = OpaqueColor::<Lab>::new([60., 4., -17.]);
     ///
-    /// assert!(lighter.difference(expected) < 0.001);
+    /// assert!(lighter.difference(expected) < 1e-4);
     /// ```
     ///
     /// [Lab]: crate::Lab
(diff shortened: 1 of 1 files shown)


color/src/color.rs
```diff
@@ -404,8 +397,7 @@ impl<CS: ColorSpace> AlphaColor<CS> {
 
     /// Map the hue of the color.
     ///
-    /// In a color space that naturally has a hue component, map that value.
-    /// Otherwise, do the mapping in [Oklch]. The hue is in degrees.
+    /// See [`OpaqueColor::map_hue`] for more details.
     #[must_use]
     pub fn map_hue(self, f: impl Fn(f32) -> f32) -> Self {
```
DJMcNab, line 402: This is orthogonal, but is there some name we could give this function to hint to the user that the value is in degrees? (i.e. in inlay hints).

DJMcNab (commented):
I'm conflicted by this; on the one hand, having to repeat the same docs is really not good. But on the other hand:
- the actual implementation is repeated, so any change should be reflected in both implementations
- this code shouldn't be *too* liable to changing.

tomcur:
That's fair. We could also link to `OpaqueColor` for the examples instead, or duplicate the examples.

raphlinus (approved):
I agree there are tradeoffs, but this seems like an improvement, so I'm fine with it going in.

DJMcNab:
I'm not sure it is an improvement in that it turns useful docs into a pointer. But I'm not going to block it. Normalisation isn't an unreasonable goal

waywardmonkeys:
They won't be changing a lot, so I don't mind the duplication. It isn't like we'll be editing them a lot and having to keep them in sync.

And that's true even when there are doc tests ... which would be by their nature testing the separate structs / code paths, so that's good "duplication".

tomcur:
As an alternative, I've removed the links and completed the docs with examples.

DJMcNab (approved):

waywardmonkeys (approved):

tomcur:
Woops, should've updated the title.

PR #57, opened 2024-11-22
Add conversion roundtrip tests

tomcur:
This tests roundtripping equality of all the rectangular color spaces. The precision bound is relatively tight: currently arbitrarily chosen as 16 float epsilons. This allows doing many tests without manually specifying the expected values.

I've added a test for Oklab as well, but all color spaces can be added in time. (I'm out of time for today.) We probably do need to think a little bit about the test cases that are interesting to hit. This now tests somewhat arbitrary values, but it would be worthwhile to hit component values near the piece edges of transfer functions, or near the hue edges, etc.


color/src/colorspace.rs
```diff
@@ -1139,10 +1139,81 @@ impl ColorSpace for Hwb {
...
+
+            for color in colors {
+                let intermediate = Source::convert::<Dest>(*color);
+                let roundtripped = Dest::convert::<Source>(intermediate);
+
+                // The roundtrip error is measured in linear sRGB. This adds more conversions, but
+                // makes the components analogous.
+                let linsrgb_color = Source::to_linear_srgb(*color);
+                let linsrgb_intermediate = Dest::to_linear_srgb(intermediate);
+                let linsrgb_roundtripped = Source::to_linear_srgb(roundtripped);
```
tomcur, line 1170: It feels like these additional conversions could be dropped somehow, but I am not sure yet how. The issue is that a large magnitude in one component can cause an error relative to that magnitude in another component, but because in arbitrary color spaces the components can have quite different scales and meanings, it becomes difficult to test to a tight bound.

DJMcNab: Would it be feasible to have a budget of floating point numbers in between? So for example, there are 13 floats between 100 and 100.0001 (according to `100.0001f32.to_bits() - 100.0f32.to_bits()` [stackoverflow](https://stackoverflow.com/a/3588000)).

I don't know if that's exactly what you want, but just adding it as an option.

waywardmonkeys:
Would you be able to rebase this after my other changes today that are in conflict with this?

tomcur:
> Would you be able to rebase this after my other changes today that are in conflict with this?

Yes, I have time tomorrow.

color/src/colorspace.rs
tomcur: In general that's a neat way to measure to tight bounds, I've heard that called "units in last place" before. Maybe we can use it elsewhere for our tests. I don't think it'll work for this specific test, as it's the maximal magnitude of the components that can influence the error in all components.

For example, roundtripping linear srgb (1, 0, 0) to XYZ-D65 and back using our implementation, you get an error on the order of 10^-8 in the green channel. Do the same with linear srgb (100, 0, 0) (very out-of-gamut), and you get an error on the order of 10^-6. (In this case, both errors are well below one relative float epsilon).

DJMcNab: Right, because it's doing big matrix multiplies. That makes sense

DJMcNab (approved):
I can't comment on how useful or meaningful these tests are, but we can always rip them out again if they're not good, and they don't impact the public API.

tomcur:
I think this test can be superseded when all color spaces have tests specified to tight bounds. In the meantime this increases confidence in our correctness.

PR #61, opened 2024-11-23
Optionally support `serde`.

waywardmonkeys:
This adds support for `serde::Deserialize` and `serde::Serialize` using `serde-derive` for `AlphaColor`, `OpaqueColor`, `PremulColor`, `Rgba8`.

This does not provide it for `DynamicColor` to avoid dealing with the non-exhaustive `ColorSpaceTag`.

It uses `derive` rather than hand-implementing the traits as the latter is very verbose and most things using `serde` will already have enabled the `derive` feature. (This is also true for the other Linebender crates.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This release has an [MSRV][] of 1.82.
 ### Added
 
 * Add `BLACK`, `WHITE`, and `TRANSPARENT` constants to the color types. ([#64][] by [@waywardmonkeys][])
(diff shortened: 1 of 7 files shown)


waywardmonkeys:
The fact that this uses the `derive` feature for `serde` doesn't change the decision to not use it for `bytemuck`.

In particular, crates like `resvg`, `tiny-skia`, and those from Makepad do use `bytemuck`, don't use `serde` and very much do not want a dependency on the whole proc macro world that `serde-derive` builds on.

raphlinus (approved):
Looks straightforward, and I agree with the decision not to hand-derive. People who use serde are opting in to that whole thing.

My personal feeling is that if we're annotating these types with serde, then `DynamicColor` belongs also, but I understand the concern about non-exhaustive and can defer that discussion to later.

PR #64, opened 2024-11-24
Add `BLACK`, `WHITE`, `TRANSPARENT` constants to color types.

### LINKED ISSUE #63: Add more color constants?
When converting Vello and Peniko to use this crate, there are a few constants that get referenced that would be nicer to have at a different level of usage:

* `BLACK` could be defined for each of the color types as the components being `0.`
* `WHITE` could be defined on each `ColorSpace` and let the color types have `const WHITE = CS::WHITE`
* `TRANSPARENT` could be defined for each of the color types that have an alpha

These are generally useful constants and don't need to be obtained via `color::palette::css:BLACK` and so on.

Also, by doing the per-colorspace white, we'd have something that doesn't otherwise exist yet in the crate.

waywardmonkeys:
These are useful values to have without having to go to the palette. Additionally, this allows for getting the correct value for the colorspace for white without having to go through a conversion.

Fixes #63.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,19 @@ You can find its changes [documented below](#010-2024-11-20).
 
 This release has an [MSRV][] of 1.82.
 
+### Added
+
+* Add `BLACK`, `WHITE`, and `TRANSPARENT` constants to the color types. ([#64][] by [@waywardmonkeys][])
+
 ## [0.1.0][] (2024-11-20)
 
 This release has an [MSRV][] of 1.82.
 
 This is the initial release.
+
+[@waywardmonkeys]: https://github.com/waywardmonkeys
+
+[#64]: https://github.com/linebender/color/pull/64
 
 [Unreleased]: https://github.com/linebender/color/compare/v0.1.0...HEAD
 [0.1.0]: https://github.com/linebender/color/releases/tag/v0.1.0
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
This is one of the changes that will help make using `color` in `peniko` and `vello` a bit easier.

raphlinus (approved):
Looks great, thanks! I agree this is useful to have.

CHANGELOG.md
```diff
@@ -15,12 +15,20 @@ You can find its changes [documented below](#010-2024-11-20).
 
 This release has an [MSRV][] of 1.82.
 
+### Added
+
+* Add `BLACK`, `WHITE`, and `TRANSPARENT` constants to the color types. ([#64][] by [@waywardmonkeys][])
```
DJMcNab, line 20: Is it worth noting that this is a breaking change?

It feels like there's an interesting thing here, that ideally we'd have different definitions of breaking for creators of custom colorspaces and users of the crate - creators of custom colorspaces are *so* rare that kind of mad.

I guess color would probably be a good use-case for the semver-trick.

waywardmonkeys: There will be some other breaking changes as well. But I guess we could call out that `ColorSpace` must have a `WHITE_COMPONENTS` const now?

DJMcNab: Yeah, that is the breaking change. Our standards for marking breaking changes aren't very developed.

PR #66, opened 2024-11-24
Add `PremulRgba8`, `Rgba8::to_u32`

waywardmonkeys:
Packing a `PremulColor` to 8 bit channels is a common operation and is done in Vello, so we now have support for a `PremulRgb8a` and some `to_u32` operations as well.

### DIFF
--- a/color/src/lib.rs
+++ b/color/src/lib.rs
@@ -35,8 +35,9 @@
 //!   * Appearance models and other color science not needed for rendering.
 //!   * Quantizing and packing to lower bit depths.
 //!
-//! The [`Rgba8`] type is a partial exception to this last item, as that representation
-//! is ubiquitous and requires special logic for serializing to maximize compatibility.
+//! The [`Rgba8`] and [`PremulRgba8`] types are a partial exception to this last item, as
+//! those representation are ubiquitous and requires special logic for serializing to
+//! maximize compatibility.
 //!
 //! Some of these capabilities may be added as other crates within the `color` repository,
 //! and we will also facilitate interoperability with other color crates in the Rust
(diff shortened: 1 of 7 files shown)


color/src/lib.rs
```diff
@@ -141,6 +142,21 @@ impl AlphaColor<Srgb> {
     }
 }
 
+impl PremulColor<Srgb> {
+    /// Create a color from pre-multiplied 8-bit rgba values.
+    ///
+    /// Note: for conversion from the [`PremulRgba8`] type, just use the `From` trait.
+    pub const fn from_rgba8(r: u8, g: u8, b: u8, a: u8) -> Self {
+        let components = [
+            u8_to_f32(r as u32),
```
raphlinus, line 151: If you like, you can change the signature of `u8_to_f32` to take a u8; the current signature is no doubt a leftover from code that was also unpacking 32 bit values. But that doesn't have to be this PR.

color/src/rgba8.rs
```diff
@@ -24,8 +24,49 @@ pub struct Rgba8 {
     pub a: u8,
 }
 
+impl Rgba8 {
+    /// Returns the color as a packed value.
```
raphlinus, line 28: It would be good to document that `r` is the most significant byte and `a` the least. Also I think this can and should be const.

color/src/rgba8.rs
```diff
@@ -24,8 +24,49 @@ pub struct Rgba8 {
...
+    /// Blue component.
+    pub b: u8,
+    /// Alpha component.
+    pub a: u8,
+}
+
+impl PremulRgba8 {
+    /// Returns the color as a packed value.
+    #[must_use]
+    pub fn to_u32(self) -> u32 {
```
raphlinus, line 63: const, same doc concern as above.

raphlinus (approved):
Generally looks good, some suggestions inline.

color/src/lib.rs
waywardmonkeys: Done, and that let it fit on a single line.

color/src/rgba8.rs
```diff
@@ -24,8 +24,51 @@ pub struct Rgba8 {
     pub a: u8,
 }
 
+impl Rgba8 {
+    /// Returns the color as a packed value, with `r` as the most significant byte and
+    /// `a` the least.
+    #[must_use]
+    pub const fn to_u32(self) -> u32 {
+        ((self.r as u32) << 24) | ((self.g as u32) << 16) | ((self.b as u32) << 8) | self.a as u32
```
DJMcNab, line 32: Should this use [from_be_bytes](https://doc.rust-lang.org/std/primitive.u32.html#method.from_be_bytes) instead? That is also const.

PR #67, opened 2024-11-25
Add `AlphaColor::with_alpha`

waywardmonkeys:
This is similar to `OpaqueColor::with_alpha`, but allows modifying the alpha channel on an `AlphaColor`.

This can be done via `map`, but this is `const`, so it can result in better code in scenarios like:

```
    let c = color::palette::css::GOLDENROD.with_alpha(0.5);
```


waywardmonkeys:
This grew out of making Vello use `color` and there were 2 places that were directly setting the alpha.

My initial port was using `map` to do the modification, but this feels like a more elegant way and allows it to be `const`, which `map` does not.

raphlinus (approved):

PR #68, opened 2024-11-25
Add missing asserts in white components tests

waywardmonkeys:


raphlinus (approved):
Huh, this seems like something a lint might have caught.

waywardmonkeys:
Ah it would if the helper were `#[must_use]`. Will do that here in this and re-use the existing approval. :)

PR #69, opened 2024-11-25
Add a color parse function consuming a string prefix

### LINKED ISSUE #60: Parsing may want to indicate how much of the string was consumed
This is useful when we're calling this from another parser, like in svgtypes.

tomcur:
The unparsed remainder of the string is indicated in the return value.

Resolves #60.

The existing function `parse_color` now returns
`ParseError::ExpectedEndOfString` if there are trailing characters. Some choices could be made differently:

- For `#f00fa` the error `ParseError::WrongNumberOfHexDigits` is returned. It could instead parse as `#f00f` with trailing `a`. The current implementation is consistent with parsing `ident` to the longest string and then checking for matches.
- This drops `ParseError::InvalidHexDigit`, as it is ambiguous whether the character was intended as a hex digit or is the start of the trailing string. `#fffffftr` is parsed as `#ffffff` with trailing `tr`. This causes a slightly worse error for the full-string-consuming `parse_color`, namely `ParseError::ExpectedEndOfString`. We could have `parse_color` still return `ParseError::InvalidHexDigit` for that case, but then the two functions have slightly different parsing behavior, which is not ideal (the prefix parse might succeed, whereas the full parse fails with an error different than `ExpectedEndOfString`).

### DIFF
--- a/color/src/parse.rs
+++ b/color/src/parse.rs
@@ -39,8 +39,8 @@ pub enum ParseError {
     ExpectedColorSpaceIdentifier,
     /// Expected comma
     ExpectedComma,
-    /// Invalid hex digit
-    InvalidHexDigit,
+    /// Expected end of string
+    ExpectedEndOfString,
     /// Wrong number of hex digits
     WrongNumberOfHexDigits,
 }
@@ -61,7 +61,7 @@ impl fmt::Display for ParseError {
             Self::ExpectedClosingParenthesis => "expected closing parenthesis",
             Self::ExpectedColorSpaceIdentifier => "expected color space identifier",
             Self::ExpectedComma => "expected comma",
-            Self::InvalidHexDigit => "invalid hex digit",
+            Self::ExpectedEndOfString => "expected end of string",
             Self::WrongNumberOfHexDigits => "wrong number of hex digits",
         };
         f.write_str(msg)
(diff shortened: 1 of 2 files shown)


color/src/parse.rs
```diff
@@ -451,21 +451,24 @@ impl<'a> Parser<'a> {
...
+///
+/// Returns the byte offset of the unparsed remainder of the string and the parsed color. See also
+/// [`parse_color`].
 ///
 /// # Errors
 ///
 /// Tries to return a suitable error for any invalid string, but may be
 /// a little lax on some details.
-pub fn parse_color(s: &str) -> Result<DynamicColor, ParseError> {
+pub fn parse_color_with_remainder(s: &str) -> Result<(usize, DynamicColor), ParseError> {
```
tomcur, line 463: Or `fn parse_color_prefix`, or another name.

color/src/parse.rs
```diff
@@ -482,43 +485,70 @@ pub fn parse_color(s: &str) -> Result<DynamicColor, ParseError> {
...
+        if let Ok(h) = hex_from_ascii_byte(hex_str.as_bytes()[i]) {
+            hex[i] = h;
+            i += 1;
+        } else {
+            break;
+        }
     }
-    Ok(four_bit_channels)
+
+    let four_bit_channels = match i {
```
DJMcNab, line 533: Can this use slice destructuring matching? I.e.
```
match &hex[..i] {
    [r, g, b] => [r, r, g, g, b, b, 15, 15],
// ...
    _ if hex.len() == 8 => hex,
 _ => return Err(...)
}
```

DJMcNab (commented):
We should link to the relevant spec sections, especially regarding the handling of trailing items after `#rrggbbaa`.

color/src/parse.rs
tomcur: I don't think we can:

```
533 |     let four_bit_channels = match &hex[..i] {
    |                                       ^^^^^
    |
    = note: calls in constant functions are limited to constant functions, tuple structs and tuple variants
```

tomcur: (This function doesn't _have_ to be const, it just was, and I've kept it that way.)

tomcur:
> We should link to the relevant spec sections, especially regarding the handling of trailing items after `#rrggbbaa`.

I'm not aware of an authoritative source on this, it's possible one exists, or perhaps there's prior art we can follow.

color/src/parse.rs
DJMcNab: :(

DJMcNab:
Hmm. I presumed that the css specs would specify this. I can't imagine that e.g. `border: #ffff00ff10px solid`  would be valid for a 10 pixel border. Although admittedly, I've not tried it.

color/src/parse.rs
raphlinus: I like the shorter name, I think it's clear enough.

color/src/parse.rs
raphlinus: Yeah, unfortunately this code is not as nice as it could be because of const. I do think that's nice to have though.

color/src/parse.rs
```diff
@@ -482,43 +485,70 @@ pub fn parse_color(s: &str) -> Result<DynamicColor, ParseError> {
...
+/// [`ExpectedEndOfString`](ParseError::ExpectedEndOfString) parse error. See also
+/// [`parse_color_with_remainder`].
+///
+/// # Errors
+///
+/// Tries to return a suitable error for any invalid string, but may be
+/// a little lax on some details.
+pub fn parse_color(s: &str) -> Result<DynamicColor, ParseError> {
+    let (ix, color) = parse_color_with_remainder(s)?;
+
```
raphlinus, line 508: I think it would be nice to skip trailing whitespace.

Of course, there's the related question of whether we should also skip leading whitespace. Currently we don't, but that would be easy enough.

raphlinus (approved):
I think this meets the requirements, some suggestions inline.

I continue to wonder whether `parse_color` should be a `FromStr` impl. Obviously the prefix version can't be.

PR #70, opened 2024-11-26
Impl `serde` for `DynamicColor`

waywardmonkeys:
Using `DynamicColor` for color stops in Peniko will require this (when Peniko's `serde` feature is enabled).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ This release has an [MSRV][] of 1.82.
 ### Added
 
 * Add `BLACK`, `WHITE`, and `TRANSPARENT` constants to the color types. ([#64][] by [@waywardmonkeys][])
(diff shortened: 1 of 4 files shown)


DJMcNab (approved):
This doesn't expose any new capabilities in terms of creating values, so it's probably fine.

I must admit, I'm not massively convinced that this is the serialisation format we'll want in the long-term. Can we document that the serialisation/deserialisation format isn't stable between `color` versions?

waywardmonkeys:
This is here so that things that enable `serde` in `peniko` will continue to work.

I don't think anyone should rely on `serde` remaining stable across breaking releases.

I'm also not really sure how great using `serde` on `peniko` (or `color`) would be ... just trying to keep the same set of things working.

waywardmonkeys:
I mean that `peniko` currently has a `serde` feature (as does `kurbo`), but I don't really know how or if that's used out in the wild. I could ask @ratmice!

PR #71, opened 2024-11-26
Add `with_alpha` and `multiply_alpha` on `DynamicColor`

waywardmonkeys:
If the alpha channel is missing, these operations will do nothing.

Additionally, mark some functions on `Missing` as `const`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This release has an [MSRV][] of 1.82.
 * A new `PremulRgba8` type mirrors `Rgba8`, but for `PremulColor`. ([#66][] by [@waywardmonkeys][])
 * `AlphaColor::with_alpha` allows setting the alpha channel. ([#67][] by [@waywardmonkeys][])
 * Support for the `ACEScg` colorspace. ([#54][] by [@MightyBurger][])
+* `DynamicColor` gets `with_alpha` and `multiply_alpha`. ([#71][] by [@waywardmonkeys][])
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
For `DynamicColor::multiply_alpha`, I think that if the alpha is missing (and therefore `0.0`), then the alpha remains missing.

But the interaction between that and `DynamicColor::premultiply_split` is not clear to me, since it treats a missing alpha as being opaque:

    fn premultiply_split(self) -> ([f32; 3], f32) {
        // Reference: §12.3 of Color 4 spec
        let (opaque, alpha) = split_alpha(self.components);
        let premul = if alpha == 1.0 || self.missing.contains(3) {
            opaque
        } else {
            self.cs.layout().scale(opaque, alpha)
        };
        (premul, alpha)
    }
```

So for the purposes of `multiply_alpha`, would we set it to not be missing and then set the alpha to the factor?

Same question for `with_alpha` where we're setting the alpha. Would we leave it missing or set it and clear the missing flag?

There's no code that clears missing flags, so maybe I'm wrong here and we should document that both things effectively do nothing when the alpha is explicitly missing?

waywardmonkeys:
We need `multiply_alpha` for supporting using `DynamicColor` with `peniko::ColorStop` / `peniko::Gradient`.

waywardmonkeys:
I've answered my own question for now by leaving the color unchanged when modifying the alpha channel if the alpha channel is missing.

This is what would have happened if one had done the same operation via `DynamicColor::map`.

raphlinus (approved):
Looks good, thanks!

PR #76, opened 2024-12-01
Derive `PartialEq` for `AlphaColor`, `OpaqueColor`, `PremulColor`

waywardmonkeys:


waywardmonkeys:
A possible issue here is that `DynamicColor` has a manual impl for `PartialEq` where it directly does a comparison of the bits.

raphlinus (approved):
Riiight, that's the reason I didn't do this before.

I think we need a `PartialEq`, so I'm approving this. What's far less clear is what semantics it should have. In particular, I don't think we should guarantee that a color component of 0.0 is equal to a color component of -0.0, ie reserve the right to make that change. Of course we're making no stability guarantees at this point.

tomcur:
I wonder whether it makes sense to follow #75 and `impl PartialEq`, `Hash` based on the bits. There is some divergence between the types, with `DynamicColor` not being `bytemuck::Pod`, so maybe that informs the decision. An alternative is to provide the required functionality on `DynamicColor` (maybe `to_bits`) so users can themselves implement `Hash`, `Eq` on some wrapping type.

tomcur (approved):
(Happy for this to land as-is.)

waywardmonkeys:
Landed this as is ... I need to do some stuff for `Default` as well so that we can use this with Parley which requires that a `Brush` impl `PartialEq` and `Default`.

PR #77, opened 2024-12-01
`(Premul)Rgba8::to_u32` now returns in little endian format.

waywardmonkeys:
This matches the WebGPU spec which states that the internal layout of values is such that values in host-shared buffers are stored in little-endian format: https://www.w3.org/TR/WGSL/#internal-value-layout


raphlinus (approved):
Agreed, I think it makes sense to standardize on this ordering. Thanks!

I am wondering about conversions into Rgba8, but I think that's less urgent. I'm also wondering (out loud) whether the `From` trait might be appropriate here.

DJMcNab (commented):
This raises questions about the bytemuck implementations. Endianness is like affinity, in that it's really hard to reason about quickly, but I think the bytemuck implementation is big endian? Should we re-order the fields to make that also be little endian?

We should have a test that they result in the same value, at least.

waywardmonkeys:
I would be more inclined to remove that particular bytemuck impl in favor of using the new to_u8__array and the other stdlib features.

PR #78, opened 2024-12-04
Fix stray closing paren in hex serialization

raphlinus:
I was printing debug colors and noticed an extraneous parenthesis. Also adds test.

PR #79, opened 2024-12-04
`HueDirection` now impls `PartialEq`

waywardmonkeys:
This is needed in Peniko when a `Gradient` has a `HueDirection` member.


waywardmonkeys:
I also moved the changelog entry for #78 as was requested in that issue.

DJMcNab (approved):

PR #80, opened 2024-12-05
Add `serde` for `HueDirection`

waywardmonkeys:
This is needed for adding `HueDirection` to `Gradient` in Peniko.


tomcur (approved):
Same concern about non-exhaustive here as on `DynamicColor`.

PR #81, opened 2024-12-05
Add `bytemuck` support for `HueDirection`, and `ColorSpaceTag`

waywardmonkeys:
These now impl `CheckedBitPattern`, `Contiguous`, `NoUninit`, and `Zeroable`. They do not impl `Pod` as they are only valid for a subset of the bit patterns.

Additionally, these types are now `repr(u8)` and have directly assigned integer values so that we can be sure that the impl of the `Contiguous` trait is correct.

This will likely be useful when working with gradients in Vello and it has to encode things to the draw stream.

It may also be useful when we're doing a C FFI in the future.

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -89,17 +89,19 @@ pub struct PremulColor<CS> {
 #[derive(Clone, Copy, Default, Debug, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[non_exhaustive]
+#[repr(u8)]
 pub enum HueDirection {
     /// Hue angles take the shorter of the two arcs between starting and ending values.
     #[default]
-    Shorter,
+    Shorter = 0,
     /// Hue angles take the longer of the two arcs between starting and ending values.
-    Longer,
+    Longer = 1,
     /// Hue angles increase as they are interpolated.
-    Increasing,
+    Increasing = 2,
     /// Hue angles decrease as they are interpolated.
-    Decreasing,
+    Decreasing = 3,
     // It's possible we'll add "raw"; color.js has it.
+    // NOTICE: If a new value is added, be sure to add modify `MAX_VALUE` in the bytemuck impl.
 }
 
 /// Fixup hue based on specified hue direction.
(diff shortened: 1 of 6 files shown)


color/src/color.rs
```diff
@@ -89,17 +89,19 @@ pub struct PremulColor<CS> {
...
-    Longer,
+    Longer = 1,
     /// Hue angles increase as they are interpolated.
-    Increasing,
+    Increasing = 2,
     /// Hue angles decrease as they are interpolated.
-    Decreasing,
+    Decreasing = 3,
     // It's possible we'll add "raw"; color.js has it.
+    // NOTICE: If a new value is added, be sure to add modify `MAX_VALUE` in the bytemuck impl.
```
DJMcNab, line 104: https://doc.rust-lang.org/stable/core/ptr/fn.read.html is const stable, so it *might* be possible to write a test which fails to compile if this is correct (i.e. using a compile_fail doctest). I don't have time at the moment to prototype this.

DJMcNab (approved):
Rocket science idea for testing, doesn't need to block this PR. I trust these impls to be correct.

tomcur (approved):

color/src/tag.rs
```diff
@@ -20,37 +20,39 @@ use crate::{
...
+    A98Rgb = 9,
     /// The [`ProphotoRgb`] color space.
-    ProphotoRgb,
+    ProphotoRgb = 10,
     /// The [`Rec2020`] color space.
-    Rec2020,
+    Rec2020 = 11,
     /// The [`AcesCg`] color space.
-    AcesCg,
+    AcesCg = 12,
```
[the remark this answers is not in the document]
waywardmonkeys: Sure, feel free to do a PR that does that! :)

color/src/color.rs
waywardmonkeys: I'm dumb so I don't even have a clue what you mean.

waywardmonkeys: Yeah, if only Rust had a real reflection system on the way.

DJMcNab:
```rust
#[repr(u8)]
enum MyEnum {
    A = 0,
    B = 1
}

const _: MyEnum = {
    let val = 1u8;
    unsafe {
        core::ptr::read((&raw const val).cast())
    }
};
```
compiles, whereas:
```rust
const _: MyEnum = {
    let val = 2u8;
    unsafe {
        core::ptr::read((&raw const val).cast())
    }
};
```
Fails to compile. This is on latest stable.

PR #82, opened 2024-12-06
Validate the `Contiguous` impls

DJMcNab:
As discussed in https://github.com/linebender/color/pull/81#discussion_r1871613918

### DIFF
--- a/color/src/impl_bytemuck.rs
+++ b/color/src/impl_bytemuck.rs
@@ -103,7 +103,7 @@ mod tests {
         AlphaColor, ColorSpaceTag, HueDirection, OpaqueColor, PremulColor, PremulRgba8, Rgba8, Srgb,
     };
     use bytemuck::{checked::try_from_bytes, Contiguous, TransparentWrapper, Zeroable};
-    use core::marker::PhantomData;
+    use core::{marker::PhantomData, ptr};
 
     fn assert_is_pod(_pod: impl bytemuck::Pod) {}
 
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
We'd need it to validate that the `MAX_VALUE` + 1 (if less than the current type size's max value) is not valid. That's the case that is most likely to fail ... when someone adds a variant and forgets to update.

And at that point, it would be better if this were generic and not copy/paste for each type as I have changes in Peniko that do this for 3 more types and will later have another 3-5 enums to do ...

tomcur (approved):
Very cool that this is possible. To me it seems like a reasonable trade-off between code complexity and having confidence the unsafe impls remain correct. When [`std::mem::variant_count`](https://doc.rust-lang.org/std/mem/fn.variant_count.html) becomes available on stable, all necessary conditions could be tested.

tomcur:
> We'd need it to validate that the `MAX_VALUE` + 1 (if less than the current type size's max value) is not valid.

The `compile_fail` doctests do that, it's quite neat.

waywardmonkeys:
oh, now that I'm actually thinking and not distracted by chores, what I asked for isn't possible, I guess until that new thing is stable like @tomcur said.

DJMcNab:
> And at that point, it would be better if this were generic and not copy/paste for each type as I have changes in Peniko that do this for 3 more types and will later have another 3-5 enums to do ...

That would require it to be a macro - a macro creating doctests should be possible, but it would be extremely ugly.

PR #85, opened 2024-12-10
Default color types to opaque white

waywardmonkeys:
We choose an opaque color because otherwise the `OpaqueColor` would have a substantially different default value than the other color types.

White is the same as the default color within `bevy_color`.

The default color is used by the Brush in Parley and likely other places.

### DIFF
--- a/color/src/dynamic.rs
+++ b/color/src/dynamic.rs
@@ -6,6 +6,7 @@
 use crate::{
     color::{add_alpha, fixup_hues_for_interpolate, split_alpha},
     AlphaColor, ColorSpace, ColorSpaceLayout, ColorSpaceTag, HueDirection, LinearSrgb, Missing,
+    Srgb,
 };
 use core::hash::{Hash, Hasher};
 
@@ -24,6 +25,9 @@ use core::hash::{Hash, Hasher};
 /// In other contexts, missing colors are interpreted as a zero value.
 /// When manipulating components directly, setting them nonzero when the
 /// corresponding missing flag is set may yield unexpected results.
+///
+/// The default value is an opaque white in the [sRGB](Srgb) color space with
+/// no missing components. We don't recommend relying upon this default.
 ///
 /// [Oklch]: crate::Oklch
 #[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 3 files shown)


color/src/dynamic.rs
```diff
@@ -365,6 +369,16 @@ impl DynamicColor {
     }
 }
 
+impl Default for DynamicColor {
+    fn default() -> Self {
+        Self {
```
raphlinus, line 374: I think this would read more smoothly as `Self::from_alpha_color(AlphaColor::<Srgb>::default())` (or WHITE if we wanted to be explicit, I'm a bit torn).

raphlinus (approved):
I'm ok with this, though I believe the canonical default is [mediumorchid](https://bikeshed.com/).

waywardmonkeys:
I'm going to keep this approval in my pocket for now and see if we can avoid needing it! :)

color/src/dynamic.rs
waywardmonkeys: Done.

DJMcNab:
In our current world without https://github.com/rust-lang/rust/issues/132162 on stable, I think having this is *fine*. I'd personally still advocate for a vile default, so that users can notice not overriding it, but not enough to block the "safe" white.

tomcur (approved):
Doing a vile default is a little bit tricky, and may require specifying the color on the color space: doing a conversion at runtime from some vile color in a specific color space is quite a bit of operations for a color that is likely to be overwritten, and it may fall outside that color space's natural gamut, making it a bit messy.

color/src/dynamic.rs
```diff
@@ -25,6 +26,9 @@ use core::hash::{Hash, Hasher};
 /// When manipulating components directly, setting them nonzero when the
 /// corresponding missing flag is set may yield unexpected results.
 ///
+/// The default value is an opaque white in the [sRGB](Srgb) colorspace with
+/// no missing components. We don't recommend relying upon this default.
+///
```
tomcur, line 31: Missing space in "colorspace" and a trailing newline

waywardmonkeys:
Many color libraries don’t provide a default… I am fixing the things that need it for now and my inclination is to only merge this under duress.

color/src/dynamic.rs
waywardmonkeys: Fixed. (Just in case we do merge it.)

DJMcNab: Can we add a typos config for this?

waywardmonkeys: @DJMcNab Not that I know of ... sure, we can add something to the config, but then it'll warn on `colorspace` as a module name. If you add something to both the identifiers and the words, then you don't get warnings.

DJMcNab: Maybe that's not a false-positive?

If we're calling it color space everywhere, maybe the module should be `color_space`?

waywardmonkeys:
Converting to draft as there is no intent to merge at this time.

waywardmonkeys:
I hope to never need this.

PR #86, opened 2024-12-10
Derive PartialEq by hand to avoid ColorSpace: PartialEq bound

waywardmonkeys:
There's no need for each `ColorSpace` to have to impl `PartialEq` since they're zero-sized types.


raphlinus (approved):
Yup I think this is fine. I note it doesn't impl for other colorspaces, but I think it's good for that to be a type error rather than us trying to solve the philosophical question of whether colors in different colorspaces can be equal.

PR #92, opened 2024-12-12
`CacheKey` type for maps which use colours as keys

DJMcNab:
This will be useful for Vello.

This is done through two new traits, `BitEq` and `BitHash`. These are used in the `Eq` and `Hash` implementations of `CacheKey<T>`.
These traits are implemented for f32, and the colour types in this crate (except the 8 bit types)

### DIFF
--- a/color/src/cache_key.rs
+++ b/color/src/cache_key.rs
@@ -23,11 +23,26 @@
 use core::hash::{Hash, Hasher};
 
 /// A key usable in a hashmap to compare the bit representation
-/// types containing colours.
+/// types containing colors.
 ///
 /// See the [module level docs](self) for more information.
 #[derive(Debug, Copy, Clone)]
+#[repr(transparent)]
 pub struct CacheKey<T>(pub T);
+
+impl<T> CacheKey<T> {
+    /// Create a new `CacheKey`.
+    ///
+    /// All fields are public, so the struct constructor can also be used.
+    pub fn new(value: T) -> Self {
+        Self(value)
+    }
+
+    /// Get the inner value.
+    pub fn into_inner(self) -> T {
+        self.0
+    }
+}
 
 // This module exists for these implementations:
 
@@ -35,7 +50,7 @@ pub struct CacheKey<T>(pub T);
 impl<T: BitEq> Eq for CacheKey<T> {}
 impl<T: BitEq> PartialEq for CacheKey<T> {
     fn eq(&self, other: &Self) -> bool {
-        self.bit_eq(other)
+        self.0.bit_eq(&other.0)
     }
 }
 // If we implement Eq, BitEq's implementation matches that of the hash.
@@ -45,7 +60,7 @@ impl<T: BitHash> Hash for CacheKey<T> {
     }
 }
 
-/// An hash implementation for types normally wouldn't have one,
+/// A hash implementation for types which normally wouldn't have one,
 /// implemented using a hash of the bitwise equivalent types when needed.
 ///
 /// If a type is `BitHash` and `BitEq`, then it is important that the following property holds:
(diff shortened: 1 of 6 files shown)


color/src/cache_key.rs
```diff
@@ -0,0 +1,200 @@
...
+
+/// An equivalence relation for types normally wouldn't have
+/// one, implemented using a bitwise comparison when needed.
+///
+/// See the docs on [`Eq`] for more information.
+///
+/// Useful for creating caches based on exact values.
+/// See the [module level docs](self) for more information.
+pub trait BitEq {
+    /// Returns true if `self` is the same "value" as other.
```
waywardmonkeys, line 117: Maybe say bit value?

color/src/lib.rs
```diff
@@ -84,12 +84,13 @@
...
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 
+pub mod cache_key;
 mod color;
 mod colorspace;
+mod dynamic;
 mod gradient;
 mod missing;
 // Note: this may become feature-gated; we'll decide this soon
-mod dynamic;
```
waywardmonkeys, line 92: Not sure which mid the comment above this referred to… but if it was dynamic, it should move with it?

DJMcNab: Well caught! This is a casualty of rustfmt, unsurprisingly.

color/src/cache_key.rs
```diff
@@ -0,0 +1,200 @@
...
+//! your cache.
+//! `T` is the key type (i.e. a color) which you want to use as the key.
+//! This `T` must implement both [`BitHash`] and [`BitEq`], which are
+//! versions of the standard `Hash` and `Eq` traits which support implementations
+//! for floating point numbers which might be unexpected outside of a caching context.
+
+use core::hash::{Hash, Hasher};
+
+/// A key usable in a hashmap to compare the bit representation
+/// types containing colours.
```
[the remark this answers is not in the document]
DJMcNab: Under duress...

color/src/cache_key.rs
```diff
@@ -0,0 +1,200 @@
...
+///
+/// See the [module level docs](self) for more information.
```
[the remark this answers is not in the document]
DJMcNab: Neither did I! I tried it and it worked though. It makes sense that you can link to `self::Item`.

color/src/cache_key.rs
```diff
@@ -0,0 +1,200 @@
...
+impl<T: BitHash> BitHash for &T {
+    fn bit_hash<H: Hasher>(&self, state: &mut H) {
+        T::bit_hash(*self, state);
+    }
+}
+
+// Don't BitHash tuples, not that important
+
+/// An equivalence relation for types normally wouldn't have
+/// one, implemented using a bitwise comparison when needed.
```
tomcur, line 110: Missing "that" in "for types normally wouldn't have one"

color/src/cache_key.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+impl<T: BitHash, const N: usize> BitHash for [T; N] {
+    fn bit_hash<H: Hasher>(&self, state: &mut H) {
+        self[..].bit_hash(state);
+    }
+}
+
+impl<T: BitHash> BitHash for [T] {
+    fn bit_hash<H: Hasher>(&self, state: &mut H) {
+        state.write_usize(self.len());
```
tomcur, line 94: It's not stabilized yet, but when it is, this should use [`Hasher::write_length_prefix`](https://doc.rust-lang.org/stable/std/hash/trait.Hasher.html#method.write_length_prefix)

color/src/lib.rs
DJMcNab: Actually, this isn't the first time it's been a casualty of rustfmt, see - https://github.com/linebender/color/pull/12/files#diff-4c57624c2719867d1070e79f01066369fa55f94eec03f36b4b2e0ecf410e190aR28

I'll restore it to the original place. Although we really should probably just remove the comment

color/src/dynamic.rs
```diff
@@ -365,28 +366,29 @@ impl DynamicColor {
...
+    fn bit_eq(&self, other: &Self) -> bool {
         self.cs == other.cs
             && self.missing == other.missing
-            && self.components[0].to_bits() == other.components[0].to_bits()
-            && self.components[1].to_bits() == other.components[1].to_bits()
-            && self.components[2].to_bits() == other.components[2].to_bits()
-            && self.components[3].to_bits() == other.components[3].to_bits()
+            && self.components.bit_eq(&other.components)
+    }
+}
```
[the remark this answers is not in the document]
DJMcNab: To be clear, they *probably* shouldn't cache the serialisation and instead just get it on-demand. But I guess if you're doing string interning, it might make sense?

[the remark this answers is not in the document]
tomcur: Yes, I think it makes sense. For example, when pretty-printing a style sheet, you may expect the same color to show up many times.

color/src/cache_key.rs
```diff
@@ -0,0 +1,210 @@
...
+//! - Equality on these types is not reflexive (consider [NaN](f32::NAN)).
+//! - Certain values have two representations (`-0` and `+0` are both zero).
+//!
+//! However, it is still useful to create caches which key off these values.
+//! These are caches which don't have any semantic meaning, but instead
+//! are used to avoid redundant calculations or storage.
+//!
+//! Color supports creating these caches by using [`CacheKey<T>`] as the key in
+//! your cache.
+//! `T` is the key type (i.e. a color) which you want to use as the key.
```
tomcur, line 18: maybe this "i.e." should be "e.g." as we also support slices, and users can themselves implement `BitEq`, `BitHash`.

tomcur (approved):
Looks great!

color/src/cache_key.rs
DJMcNab: This isn't really intended to be used for non-colour cases (hence the impl only being for f32 and not all the other numeric types). Like yes, you can implement this yourself, but we are being careful about that.

PR #94, opened 2024-12-13
Doc improvements

waywardmonkeys:


color/src/rgba8.rs
```diff
@@ -66,6 +83,8 @@ pub struct PremulRgba8 {
 
 impl PremulRgba8 {
     /// Returns the color as a `[u8; 4]`.
+    ///
+    /// The color values will be in the order `[r, g, b, a]`.
```
nicoburns, line 87: This kind of documentation is invaluable!

nicoburns (approved):

PR #95, opened 2024-12-15
Separate error type for parse vs `FromStr`

waywardmonkeys:
The error type `ParseError` will shortly use a lifetime to have a lifetime bound tied to the input data.

The error type for `FromStr` is not permitted to have a lifetime bound, so using the same type isn't possible.

### DIFF
--- a/color/src/lib.rs
+++ b/color/src/lib.rs
@@ -114,7 +114,7 @@ pub use colorspace::{
 pub use dynamic::{DynamicColor, Interpolator};
 pub use flags::{Flags, Missing};
 pub use gradient::{gradient, GradientIter};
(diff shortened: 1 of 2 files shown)


nicoburns:
Would it make sense to have a range of indexes in the error type instead to avoid this issue? (would also potentially be more convenient for consumers of the library if the error type isn't borrowing the input).

tomcur:
I think Nico's suggestion of using a range makes sense. Having the range is also useful for consumers if they want to print an error message where the error is pointed out in context.

waywardmonkeys:
There are problems with ranges with the current code, but not something that can't be solved well enough.

This would break the ability to do #111 and I prefer having #111 to this.

PR #96, opened 2024-12-15
Begin testing the parser components more

waywardmonkeys:


waywardmonkeys:
Before I add the index stuff to the error type, I wanted to have some tests, so separating some of that out.

tomcur (approved):

PR #97, opened 2024-12-16
Move docs on Main Types to the top

nicoburns:
Moves the "Main types" section above the "Scope and goals" section in the lib.rs docs.

Justification: I think the documentation on "what are the main types in this crate" / "how do I use this crate" ought to be front and center. Whereas the "background" on scope and goal is more secondary. The "Main Types" section is also quite a bit shorter than the "Scope and goals" section, so it means both fit "above the fold".

More concretely, I was reading the docs and trying to work out which type(s) could be considered the primary/default color representation(s). And it took me quite a bit of time to find it even though there was already a section explaining it.

### DIFF
--- a/color/README.md
+++ b/color/README.md
@@ -33,6 +33,20 @@ See https://linebender.org/blog/doc-include/ for related discussion. -->
 Color is a Rust crate which implements color space conversions, targeting at least
 [CSS Color Level 4].
 
+## Main types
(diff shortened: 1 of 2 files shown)


waywardmonkeys (approved):
I'm okay with this. You'll have to fix the readme check before landing this.

PR #98, opened 2024-12-17
Update version to 0.2.0 for publish

waywardmonkeys:


waywardmonkeys:
I'm intentionally deciding that we won't improve the `ParseError` type at this point and will write more about why soon (probably).

DJMcNab (approved):
I'm still a bit cautious about the `bytemuck` impl for Rgba8 not being consistent with the u32 conversion. But I don't think that should *necessarily* block this.

Edit: I've just tested it, and it appears to be consistent? I'm not sure why

tomcur:
Have you been able to test with a big endian system as well?

DJMcNab:
No, not yet. I think miri could?

PR #100, opened 2024-12-17
Always return little-endian packed value in Rgba8::to_u32

tomcur:
(And PremulRgba8.)

If the intention is for the bytes of R, G, B and A to always be in the same memory order, I believe this patch is required. The docstring on the method suggests that's the intention.

In the order of memory of `Rgba8::to_u8_array`, the red channel is at the lowest memory address and therefore (in little endian) the least significant byte. On a big endian system, using `from_le_bytes` would move the red byte to the highest memory address.

### DIFF
--- a/color/src/rgba8.rs
+++ b/color/src/rgba8.rs
@@ -45,7 +45,7 @@ impl Rgba8 {
     /// most significant byte and `r` the least.
     #[must_use]
     pub const fn to_u32(self) -> u32 {
-        u32::from_le_bytes(self.to_u8_array())
+        u32::from_ne_bytes(self.to_u8_array())
     }
 }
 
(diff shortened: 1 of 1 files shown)


DJMcNab (commented):
This function has gone through all three of these functions in the last couple of weeks.

We really need some miri testing for a big-endian system...

I think... that I agree that this is right.

tomcur:
The invocation

```shell
$ cargo miri test --target s390x-unknown-linux-gnu --features bytemuck -- rgba8
```

fails on main. It passes with this patch.

waywardmonkeys:
I've done a PR with Miri in CI in #101 ...

waywardmonkeys (approved):
Okay ... and we'll use Miri in CI to keep things right going forward!

waywardmonkeys:
Going to merge this now so that I can perhaps do the release now. (Breaking the usual policy...)

PR #101, opened 2024-12-17
ci: Run tests under miri on a big endian platform

waywardmonkeys:


DJMcNab (approved):
The other question is do we want to make this required?

Because it needs nightly to work, I'd be inclined to say no, but maybe we should see if it becomes a problem.

waywardmonkeys:
This needs #100 to land first.

I think we can try having this required and see what happens.

waywardmonkeys:
It is nice to run this via `cargo hack`, but this may be too slow to keep this way ... going to land this as is for now though.

DJMcNab:
I don't think `cargo hack` does anything here? We always run tests with `all_features`.
It is extremely unusual to have tests running under `cargo hack` - indeed, so unusual that I didn't even notice it when reviewing this.

PR #109, opened 2024-12-21
Make color parsing case insensitive

### LINKED ISSUE #105: Mixed and uppercase idents
The code in `color` currently only handles lowercase idents while the spec says it should be case insensitive.

This also impacts use from svgtypes which currently tests that mixed cases work.

svgtypes expects that these parse:

* `"RGB(254, 203, 231)"`
* `"RgB(254, 203, 231)"`
* `RED`
* `ReD`

raphlinus:
The spec requires matching both lowercase and uppercase. The general solution is to copy the identifier into a temporary stack allocated buffer and do the case conversion there, before matching.

Fixes #105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ You can find its changes [documented below](#020-2024-12-17).
 
 This release has an [MSRV][] of 1.82.
 
+### Fixed
+
+* Make color parsing case insensitive. ([#109][] by [@raphlinus][])
+
 ## [0.2.0][] (2024-12-17)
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Can you add a quick changelog entry? It'll probably conflict with #108, but I don't mind merging this one first and then rebasing #108.

waywardmonkeys (approved):
LGTM once there's a changelog entry.

tomcur (approved):

waywardmonkeys (approved):

PR #110, opened 2024-12-21
Uncomment mixed-case serialization tests

tomcur:
These pass since https://github.com/linebender/color/pull/109.


raphlinus (approved):
Yay! More tests green.

PR #111, opened 2024-12-21
Add `FromStr` for `DynamicColor` and static colors

waywardmonkeys:
This makes it easier to parse them from a string with more concise code and not requiring a separate import of `parse_color`.

### DIFF
--- a/color/src/parse.rs
+++ b/color/src/parse.rs
@@ -9,7 +9,10 @@ use core::fmt;
 use core::str;
 use core::str::FromStr;
 
-use crate::{AlphaColor, ColorSpaceTag, DynamicColor, Flags, Missing, Srgb};
+use crate::{
+    AlphaColor, ColorSpace, ColorSpaceTag, DynamicColor, Flags, Missing, OpaqueColor, PremulColor,
+    Srgb,
+};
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
If we like this idea, then I'd add tests to the PR.

waywardmonkeys:
This would improve code in Vello's `pico_svg` as well as my current patches to `svgtypes`.

nicoburns (approved):
+1 from me on the idea of `FromStr` impls

color/src/parse.rs
```diff
@@ -543,6 +545,42 @@ pub fn parse_color(s: &str) -> Result<DynamicColor, ParseError> {
...
+    }
+}
+
+impl<CS: ColorSpace> FromStr for OpaqueColor<CS> {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        parse_color(s)
+            .map(DynamicColor::to_alpha_color)
+            .map(AlphaColor::discard_alpha)
```
DJMcNab, line 570: This one I'm less sure of. If the alpha isn't fully opaque, I'd maybe... expect this to error?

PR #113, opened 2024-12-22
Replace `From<Rgba8>` for `PremulColor` with `From<PremulRgba8>`

waywardmonkeys:
Having the `From<Rgba8>` was unintentional and should have been `From<PremulRgba8>`.

A trait impl can't be deprecated with an attribute, so just note it in the doc comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This release has an [MSRV][] of 1.82.
 ### Changed
 
 * Don't enable `serde`'s `std` feature when enabling our `std` feature. ([#108][] by @waywardmonkeys][])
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Alternatively, we could keep the `From<Rgba8>` ... but I don't see much value in doing so.

If we merge this one, then I'll file an issue for the 0.3.0 milestone to remind us to remove the code.

nicoburns (approved):

nicoburns:
We could just remove this and make the next release `0.3.0`. I don't think there much cost to that?

waywardmonkeys:
I don't want to do a 0.3.0 yet ... at the least, I'd have to deal with `ParseError` changes and I want a fairly quick 0.2.1 release.

DJMcNab (commented):
I had a comment here which I failed to submit. It's not useful now, though

PR #116, opened 2024-12-27
Fix changelog links

tomcur:
And change sentence to imperative.


xStrom (approved):
Regarding imperative, I think the better way to do these is to have the section header "Added" like in the template and the individual items should omit repeating that in any form. However we're all over the place with our changelogs and we don't need to debate that right now. This PR is an improvement with the links and shouldn't be blocked.

PR #117, opened 2024-12-28
Fix `clippy::precedence` lint

waywardmonkeys:


waywardmonkeys:
This is a new default-enabled lint in nightly clippy.

raphlinus (approved):
I agree, and generally try to be explicit in new code.

Not suggesting it for this particular issue, but one thing I've found is that multiply-add is sometimes cleaner code, `r0 * 16 + r1`. It used to be that you wanted to write the shift explicitly because it was more optimized, but that was when I was a little kid, really.

PR #118, opened 2025-01-01
Serialize `XyzD65` as `xyz-d65`

waywardmonkeys:
As noted in § 14.4 of CSS Color 4 spec, colors specified in `xyz`, which is an alias of `xyz-d65`, are in the `xyz-d65` color space.

We already only have a predefined colorspace for `XyzD65` and not just `Xyz`, but were serializing it out in the shorter form.


tomcur (approved):
Ah yes, seems `xyz-d65` is the correct name.

PR #119, opened 2025-01-01
Clamp alpha during parse, other tweaks

waywardmonkeys:
* § 4.2 of the CSS Color 4 spec states that the parsed alpha value will be clamped.
* Add a new enum `Mode` to help indicate whether we're in modern or legacy parsing mode. This can be used in additional ways in subsequent commits to improve the strictness of parsing.
* Reading an alpha value no longer automatically consumes trailing white space.

### DIFF
--- a/color/src/parse.rs
+++ b/color/src/parse.rs
@@ -87,6 +87,22 @@ enum Value<'a> {
     Dimension(f64, &'a str),
 }
 
+/// Whether or not we are parsing modern or legacy mode syntax.
+#[derive(Clone, Copy, Debug, PartialEq)]
+enum Mode {
+    Legacy,
+    Modern,
+}
+
+impl Mode {
+    fn alpha_separator(self) -> u8 {
+        match self {
+            Self::Legacy => b',',
+            Self::Modern => b'/',
+        }
+    }
+}
+
 #[expect(
     clippy::cast_possible_truncation,
     reason = "deliberate choice of f32 for colors"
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
FWIW, I found this running some of the CSS tests from Simon Sapin.

color/src/parse.rs
```diff
@@ -87,6 +87,22 @@ enum Value<'a> {
...
 }
 
+/// Whether or not we are parsing modern or legacy mode syntax.
+#[derive(Clone, Copy, Debug, PartialEq)]
+enum Mode {
+    Legacy,
+    Modern,
+}
+
+impl Mode {
```
raphlinus, line 97: A fun trick you could do here is have `Mode` to be `#[repr(u8)]` and have the values be `b','` and `b'/'`, then this method is just a cast. I'm not sure if that helps with anything :)

color/src/parse.rs
```diff
@@ -333,32 +344,40 @@ impl<'a> Parser<'a> {
...
+    ///
+    /// The value may be either number or a percentage.
+    ///
+    /// The alpha value defaults to `1.0` if not present. The value will be clamped
+    /// to the range [0, 1].
+    ///
+    /// The separator will be a `'/'` in modern mode and a `','` in legacy mode.
+    /// If no separator is present, then the default value will be returned.
+    ///
+    /// Reference: § 4.2 of CSS Color 4 spec.
```
raphlinus, line 373: For completeness, it might be worth mentioning that this also parses "none".

raphlinus (approved):
Yes, this is more correct. Thanks!

I'm seeing where I went wrong. The language for `rgb()` says "Values outside these ranges are not invalid, but are clamped to the ranges defined here at parsed-value time." so I applied the alpha clamping there, but missed the broader statement in 4.2. Good catch, and a good argument for tests.

color/src/parse.rs
waywardmonkeys: I bikeshedded this so much ... I had that in an earlier version when this enum was just `OpacitySeparator`, but I didn't want that here where it might (should) control more things than just that...

PR #124, opened 2025-01-06
Add the ACES2065-1 color space

tomcur:
This color space doesn't need anything fancy and [probably is useful](https://xi.zulipchat.com/#narrow/channel/466849-color/topic/Support.20for.20ACES2065-1.2FAP0).


color/src/colorspace.rs
```diff
@@ -540,22 +540,80 @@ impl ColorSpace for Rec2020 {
...
+/// ACES2065-1 has a reference white [near D60][aceswp]; see the [XYZ-D65 color space](`XyzD65`)
+/// documentation for some background information on the meaning of "reference white."
+///
+/// See also [`AcesCg`].
+///
+/// [aces20651]: https://draftdocs.acescentral.com/specifications/encodings/aces2065-1/
+/// [smpte]: https://pub.smpte.org/doc/st2065-1/20200909-pub/st2065-1-2021.pdf
+/// [aceswp]: https://docs.acescentral.com/tb/white-point
+#[derive(Clone, Copy, Debug)]
+pub struct Aces2065_1;
```
tomcur, line 564: The underscore in the name is "correct," but also a bit awkward, as it's just a single digit trailing.

color/src/tag.rs
```diff
@@ -48,6 +48,8 @@ pub enum ColorSpaceTag {
     ProphotoRgb = 10,
     /// The [`Rec2020`] color space.
     Rec2020 = 11,
+    /// The [`Aces2065_1`] color space.
+    Aces2065_1 = 15,
```
tomcur, line 52: How do we want to keep this sorted? Renumbering may be breaking.

color/src/tag.rs
```diff
@@ -48,6 +48,8 @@ pub enum ColorSpaceTag {
...
     /// The [`AcesCg`] color space.
     AcesCg = 12,
     /// The [`XyzD50`] color space.
```
tomcur, line 55: How do we want to keep this sorted? Renumbering may be breaking.

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ You can find its changes [documented below](#022-2025-01-03).
 
 This release has an [MSRV][] of 1.82.
 
+### Added
+
+* Support for the ACES2065-1 color space. ([#124][] by [@tomcur][])
```
waywardmonkeys, line 20: Need link below. Since I too love to forget them...

waywardmonkeys (approved):

color/src/tag.rs
waywardmonkeys: I'm okay with what you did, but I think probably would be good to put a comment somewhere in there about it. I don't think re-numbering is good except (maybe) at breaking changes.

CHANGELOG.md
tomcur: Too easy to forget!

waywardmonkeys:
Given the reaction by @raphlinus  in the Zulip thread, I think this can land today!

color/src/tag.rs
DJMcNab: I tend to agree that re-numbering in a patch release would be bad. But I also think that we should renumber these when we next make a breaking change. If people are storing color spaces in a file, then I'd like to encourage them to make their own ordering.

CHANGELOG.md
DJMcNab: My settings are:
```
"markdownlint.config": {
        "no-emphasis-as-heading": false,
        "ul-indent": false,
        "no-inline-html": false,
        "first-line-h1": false,
        "no-duplicate-heading": false
    },
```

PR #128, opened 2025-01-08
Specify WHITE_COMPONENTS to machine precision

tomcur:
These two are off by 1 ULP.


waywardmonkeys (commented):
Should we tighten the error bound in the `white_components` test?

waywardmonkeys (approved):

PR #135, opened 2025-01-18
Implement `(Premul)Rgba8::from_u32` and `from_u8_array`

### LINKED ISSUE #73: `from_u32` on `Rgba8` and `PremulRgba8`
We have `to_u32` on `Rgba8` and `PremulRgba8`, added in  https://github.com/linebender/color/pull/66#discussion_r1858528969. I can foresee use cases (e.g. reading from an image file) where you have the packed representation and would like to operate on the values. Is this an in-scope feature for this crate.

Fwiw, this is already exposed through the bytemuck implementations if you have the relevant feature enabled.

tomcur:
Closes #73.

The conversions from the packed u32 are easy to get wrong when implemented manually.

(On the other hand, these methods are also easy to use wrong.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This release has an [MSRV][] of 1.82.
 
 * Support for the ACES2065-1 color space. ([#124][] by [@tomcur][])
 * A documentation example implementing `ColorSpace`. ([#130][] by [@tomcur][])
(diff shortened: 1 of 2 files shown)


raphlinus (approved):
Looks good, thanks! It's pretty cool that big-endian is tested in CI.

tomcur:
> It's pretty cool that big-endian is tested in CI.

No kidding. `to_u32` went through all three `from_{be,le,ne}_bytes` variants.

PR #136, opened 2025-01-19
Support construction from rgb8 values

waywardmonkeys:
This is a common pattern in code in Vello examples and other places where they're just constructing an opaque color from the r,g,b values and a solid alpha.


xorgy (approved):

raphlinus (approved):
Seems like a nice ergonomic improvement, thanks.

PR #137, opened 2025-01-25
Add `AlphaColor::from_kelvin` / `OpaqueColor::from_kelvin`

waywardmonkeys:
These are feature-flagged (`feature = "temperature"`).

### DIFF
--- a/color/src/temperature.rs
+++ b/color/src/temperature.rs
@@ -1,0 +1,85 @@
+// Copyright 2025 the Color Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#![expect(
+    clippy::excessive_precision,
+    reason = "We are sticking with the constants from Tanner Helland."
+)]
+
+use crate::{AlphaColor, ColorSpace, OpaqueColor};
+
+#[cfg(all(not(feature = "std"), not(test)))]
+use crate::floatfuncs::FloatFuncs;
+
+// Implementation notes:
+//
+// * These can't be const yet as `clamp`, `ln`, and `powf` are not `const`.
+
+impl<CS: ColorSpace> AlphaColor<CS> {
+    /// Convert a temperature in Kelvin to a color.
+    ///
+    /// The Kelvin temperature will be clamped to a range of 1000-40,000.
+    pub fn from_kelvin(kelvin: f32) -> Self {
+        let [r, g, b] = CS::from_linear_srgb(from_kelvin(kelvin));
+        Self::new([r, g, b, 1.])
+    }
+}
+
+impl<CS: ColorSpace> OpaqueColor<CS> {
+    /// Convert a temperature in Kelvin to a color.
+    ///
+    /// The Kelvin temperature will be clamped to a range of 1000-40,000.
+    pub fn from_kelvin(kelvin: f32) -> Self {
+        Self::new(CS::from_linear_srgb(from_kelvin(kelvin)))
+    }
+}
+
+/// Convert a Kelvin temperature into a linear RGB color.
+fn from_kelvin(kelvin: f32) -> [f32; 3] {
+    let kelvin = kelvin.clamp(1000., 40000.) / 100.;
+
+    let red = if kelvin < 66. {
+        255.
+    } else {
+        329.698_727_446 * (kelvin - 60.).powf(-0.133_204_759_2)
+    };
+
+    let green = if kelvin <= 66. {
+        99.470_802_586_1 * kelvin.ln() - 161.119_568_166_1
+    } else {
+        288.122_169_528_3 * (kelvin - 60.).powf(-0.075_514_849_2)
+    };
+
+    let blue = if kelvin >= 66. {
+        255.
+    } else if kelvin <= 19. {
+        0.
+    } else {
+        138.517_731_223_1 * (kelvin - 10.).ln() - 305.044_792_730_7
+    };
+
+    [
+        red.min(255.) / 255.,
+        green.min(255.) / 255.,
+        blue.min(255.) / 255.,
+    ]
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{ColorSpace, OpaqueColor, Srgb};
+
+    fn assert_temperature(kelvin: f32, linear_rgb: [f32; 3]) {
+        let result = OpaqueColor::<Srgb>::from_kelvin(kelvin).to_rgba8();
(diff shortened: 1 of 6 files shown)


color/src/temperature.rs
```diff
@@ -0,0 +1,85 @@
...
+        329.698_727_446 * (kelvin - 60.).powf(-0.133_204_759_2)
+    };
+
+    let green = if kelvin <= 66. {
+        99.470_802_586_1 * kelvin.ln() - 161.119_568_166_1
+    } else {
+        288.122_169_528_3 * (kelvin - 60.).powf(-0.075_514_849_2)
+    };
+
+    let blue = if kelvin >= 66. {
```
waywardmonkeys, line 53: The writing on the post about this says that blues above 6500K are 255, but the math here and in other implementations is checking against 6600K. I think this should be fixed.

I also leave things as they are in the sample code as calculating a value between 0 and 255 and then at the end dividing by 255. This should probably do better.

color/src/temperature.rs
```diff
@@ -0,0 +1,85 @@
...
+        let expected = OpaqueColor::<Srgb>::new(Srgb::from_linear_srgb(linear_rgb)).to_rgba8();
+        assert_eq!(result, expected,
+            "Failed getting color from temperature `{kelvin}`. Expected: `{expected}`. Got: `{result}`.");
+    }
+
+    #[test]
+    fn rgb_from_kelvin() {
+        // These expected values are in linear RGB
+        assert_temperature(2700., [1., 0.6538038, 0.3427666]);
+        assert_temperature(6600., [1., 1., 1.]);
```
waywardmonkeys, line 83: Need to add more tests but also need to make sure we get values from a good source of truth.

color/src/temperature.rs
```diff
@@ -0,0 +1,85 @@
...
+use crate::floatfuncs::FloatFuncs;
+
+// Implementation notes:
+//
+// * These can't be const yet as `clamp`, `ln`, and `powf` are not `const`.
+
+impl<CS: ColorSpace> AlphaColor<CS> {
+    /// Convert a temperature in Kelvin to a color.
+    ///
+    /// The Kelvin temperature will be clamped to a range of 1000-40,000.
```
waywardmonkeys, line 21: The docs should be improved here and would be better with some color swatches and a table.

Also, need to credit the original blog post and link to Wikipedia about color temperature.

tomcur: Hmm, it is suspect to me that the calculations clamp to the sRGB gamut, but go to 1000 kelvin. I believe temperatures lower than roughly 2000 kelvin are outside sRGB's natural gamut, and for e.g. 1667 kelvin, it should be roughly  `color(srgb 1.0 0.46 -0.12)`, based on the chromaticities here https://en.wikipedia.org/w/index.php?title=Planckian_locus&oldid=1245078623#Approximation.

color/src/temperature.rs
```diff
@@ -0,0 +1,85 @@
...
+impl<CS: ColorSpace> OpaqueColor<CS> {
+    /// Convert a temperature in Kelvin to a color.
+    ///
+    /// The Kelvin temperature will be clamped to a range of 1000-40,000.
+    pub fn from_kelvin(kelvin: f32) -> Self {
+        Self::new(CS::from_linear_srgb(from_kelvin(kelvin)))
+    }
+}
+
+/// Convert a Kelvin temperature into a linear RGB color.
```
tomcur, line 37: I believe this is actually calculating sRGB (i.e. non-linear), but would need to double check.

tomcur:
sRGB doesn't necessarily need to be blessed here, users working with color temperatures may be working in color spaces like Adobe RGB, ProPhoto RGB, ACEScg, …. Wanting to chromatically adjust based on the color temperature, I think you'd probably go through XYZ. Maybe it makes sense to calculate the chromaticities `(x,y)` and use XYZ (D65?) as the base color space for this; there's more prior art we can base this on then. I'm not sure yet how `Y` could be best determined, though. Some sources use the peak luminance of the target gamut (for example, Wikipedia's Color Temperature template does that https://en.wikipedia.org/wiki/Template:Color_temperature).

color/src/temperature.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl<CS: ColorSpace> OpaqueColor<CS> {
+    /// Convert a temperature in Kelvin to a color.
+    ///
+    /// The Kelvin temperature will be clamped to a range of 1000-40,000.
+    pub fn from_kelvin(kelvin: f32) -> Self {
+        Self::new(CS::from_linear_srgb(from_kelvin(kelvin)))
+    }
+}
```
tomcur, line 35: Perhaps the constructor should just be defined on the base color space used for this. If we have `OpaqueColor::<CS>::from_kelvin()`, we need to make sure the colors aren't chromatically adapted. As in, say we calculate the color temperature in sRGB as done here (sRGB having a D65 white point), and convert that to, say, ProPhoto RGB (with a D50 white point), the color is chromatically adapted, and now represents a different absolute color, with a different color temperature.

tomcur: This was discussed during office hours. The path forward is probably to put this behind a feature flag that also introduces a new wrapper type, `AbsoluteColor`, between which conversions can be done that don't do (or undo) the implicit chromatic adaptation performed by `ColorSpace::{to_linear_srgb,from_linear_srgb,convert}`. This requires storing the white point chromaticities on `ColorSpace` as an associated const (probably without a feature flag to maintain compatibility of the type between builds with different flags), and expose the white point on `DynamicColor` as well.

tomcur (commented):
The thinking in https://github.com/linebender/color/pull/137#discussion_r1935972696 is no longer current.

Absolute conversions were implemented in https://github.com/linebender/color/pull/139, which also introduced a `Chromaticity` struct in the `xy` color space (chromaticity without luminance).

Colors specified as temperatures can be converted to a `Chromaticity`. Separately, a `Chromaticity` plus a luminance can be converted to a color in a color space.

There are multiple shapes the API for converting `Chromaticity` to a color can take. Two ideas: (1) bless `XyzD65` with a constructor taking a `Chromaticity` and a luminance, (2) put a method on `Chromaticity` generic over `ColorSpace` taking a luminance parameter that converts to an absolute color in that color space (i.e., without chromatic adaptation). We can choose to defer this to a later PR, it's straightforward for clients to get an XYZ color manually.

waywardmonkeys:
Replaced by #225.

PR #139, opened 2025-02-04
Implement absolute color conversions and chromatic adaptation

tomcur:
This adds absolute color conversions and chromatic adaptation, and is, more or less, a prerequisite for #137.

Conversion methods are added to `ColorSpace`, `ColorSpaceTag`, and `DynamicColor`, that convert between color spaces while keeping the same absolute color (i.e., if that color were to be reproduced, it would be the same physical color, but a perceptually different color under the intended reference white of the color space). These methods are suffixed `_absolute`. This also adds a `chromatically_adapt` method for manual chromatic adaptation (useful for, e.g., "white balancing" pictures).

The white points are represented as CIE `xy` chromaticities. Calculation of chromatic adaptation matrices is `const` where possible.

A follow-up to this would be to manually implement `ColorSpace::{to_linear_srgb_absolute, from_linear_srgb_absolute, convert}` for the color spaces we provide. Those can in most cases just lift directly from their non-`_absolute` counterparts, with the exceptions being XYZ-D50, ACEScg and ACES2065-1. Those methods will get somewhat simpler matrices, as the `{D50, ACES}<->D65` adaptation transforms can be dropped.


color/src/colorspace.rs
```diff
@@ -1515,4 +1646,30 @@ mod tests {
             ));
         }
     }
+
+    #[test]
+    fn absolute_conversion() {
+        assert!(almost_equal::<AcesCg>(
+            Srgb::convert_absolute::<AcesCg>([0.5, 0.2, 0.4]),
+            // Calculated using colour-science (https://github.com/colour-science/colour) with
```
tomcur, line 1654: I tried adding `colour-science = "colour-science"` to `.typos.toml`'s `[default.extend-identifiers]` section, but it keeps triggering (locally, at least). I think `typos` doesn't handle hyphens, so it's just triggering on `colour = "color"`.

tomcur: Solved, thanks @DJMcNab!

PR #144, opened 2025-03-26
Add Aces2065-1 to `same_analogous`

tomcur:
This was missed in https://github.com/linebender/color/pull/124.


waywardmonkeys (commented):
Okay, but:

* The doc comment for this function should reference the section of the CSS spec that defines this.
* I think that there should be at least one test that makes sure that this behavior works, but maybe not as part of this PR.

waywardmonkeys (approved):
(Meant to mark it as approved apart from the comments.)

PR #145, opened 2025-03-28
Fix `powerless_to_missing` for `Hwb`

tomcur:
Previously all non-rectangular spaces were assumed to have saturation or chroma as second component, but this is not the case for `Hwb`.

[Color Module 4](https://www.w3.org/TR/css-color-4/#valdef-hwb-hwb) states:

> If the sum white+black is greater than or equal to 100%, it defines an
> achromatic color, i.e. a shade of gray; when converted to sRGB the R, G
> and B values are identical and have the value white / (white + black).
> ...
> Achromatic HWB colors no longer contain any hint of the chosen hue. In
> this case, the hue component is powerless.

### DIFF
--- a/color/src/dynamic.rs
+++ b/color/src/dynamic.rs
@@ -290,12 +290,22 @@ impl DynamicColor {
         // and there is some controversy in discussion threads. For example,
         // in Lab-like spaces, if L is 0 do the other components become powerless?
         const POWERLESS_EPSILON: f32 = 1e-6;
-        if self.cs.layout() != ColorSpaceLayout::Rectangular
-            && self.components[1] < POWERLESS_EPSILON
-        {
-            let mut missing = self.flags.missing();
-            self.cs.set_h_missing(&mut missing, &mut self.components);
-            self.flags.set_missing(missing);
+        match self.cs {
+            ColorSpaceTag::Lch | ColorSpaceTag::Oklch | ColorSpaceTag::Hsl
+                if self.components[1] < POWERLESS_EPSILON =>
+            {
+                let mut missing = self.flags.missing();
+                self.cs.set_h_missing(&mut missing, &mut self.components);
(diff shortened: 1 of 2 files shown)


color/src/dynamic.rs
```diff
@@ -290,12 +290,22 @@ impl DynamicColor {
...
+            ColorSpaceTag::Lch | ColorSpaceTag::Oklch | ColorSpaceTag::Hsl
+                if self.components[1] < POWERLESS_EPSILON =>
+            {
+                let mut missing = self.flags.missing();
+                self.cs.set_h_missing(&mut missing, &mut self.components);
+                self.flags.set_missing(missing);
+            }
+            ColorSpaceTag::Hwb
+                if self.components[1] + self.components[2] > 100. - 100. * POWERLESS_EPSILON =>
+            {
```
waywardmonkeys, line 303: I think this deserves a small comment here and a pointer to the appropriate section of the spec (just section reference I think, don't need a URL).

Also, would really like to see a couple of tests just to ensure that the behavior is clearly desired and checked for.

waywardmonkeys: Also, why the complicated stuff after `>` in the comparison?

tomcur: > Also, why the complicated stuff after > in the comparison?

Saturation/chroma are in the range 0-1, but HWB's whiteness and blackness are 0-100. The epsilon is tuned for the range 0-1. Doing the correction for the range behind the comparison allows Rust to calculate at compile-time, requiring only the addition on the left at runtime. An alternative would be to make another const specifically for the 0-100 range, but that's also a bit heavy-handed.

waywardmonkeys: I think I'd rather see a `const` and it can have a comment explaining it. I don' think I'd remember what you wrote here in 2 years. :)

PR #149, opened 2025-04-03
Add `Missing::EMPTY` to allow getting a `const` empty `Missing`

tomcur:
The only alternative currently is `Missing::default()`, but that's not `const`.


waywardmonkeys:
Not sure what this gets us today, but can see where it might get us more in the future.

waywardmonkeys (approved):

PR #153, opened 2025-04-16
Test implicit vs explicit chromatic adaptation

tomcur:
This is useful to test the current implementation of absolute conversions (added in https://github.com/linebender/color/pull/139), but the main motivation is to have tests in place as preparation for manual implementations of `ColorSpace::{to_linear_srgb_absolute, from_linear_srgb_absolute}` for the non-D65 color spaces.

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -1669,4 +1669,69 @@ mod tests {
             1e-4,
         ));
     }
+
+    /// Test whether `ColorSpace::convert` with implicit chromatic adaptation results in the same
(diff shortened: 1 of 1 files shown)


tomcur:
(Perhaps the test above this one is now a bit superfluous, though it doesn't do *exactly* the same thing. Probably it doesn't hurt to keep.)

tomcur (commented):
(Perhaps the test above this one is now a bit superfluous, though it doesn't do exactly the same thing. Probably it doesn't hurt to keep.)

raphlinus (approved):

Issue #106, opened 2024-12-21
Lightness out of range


waywardmonkeys:
There is a test in svgtypes for when HSL values are out of range from parse time.

It expects `hsl(800, 150%, -50%)` to result in `rgba(0, 0, 0, 255)` as the lightness is clamped to 0, and so the value is a black color.

While the CSS Color Module 4 spec says that, when parsing HSL, saturation is clamped, it does not say it about lightness.

* Section 4.2 does say to clamp alpha.
* Section 5.1 does say to clamp r, g, and b when parsing `rgb()`.
* Section 7 says to clamp saturation for HSL.
* Section 9.3 says to clamp L for Lab and Lch.
* Section 9.4 says to clamp L for OkLab and OkLch.
* Section 13.2 says that when gamut mapping to ran rgb destination, the out of range lightness values should be handled by resulting in either black or white colors.

Even if we clamp at parse time, we'd have to think also about what to do when the lightness has been modified (like via `map_lightness`) and is now out of range due to calculations.

tomcur:
I did a little bit of digging. 

1. HSL's lightness used to be clamped, but it became unbounded in https://github.com/w3c/csswg-drafts/commit/44c37787dfeeb299bad9be0f2dd7290f0f76bb53, discussion at https://github.com/w3c/csswg-drafts/issues/8444.
3. Oklab/Lab became bounded in https://github.com/w3c/csswg-drafts/commit/d504654964b10634c9f77dc53b864add4074f530, discussion at https://github.com/w3c/csswg-drafts/issues/7924.
4. The incorrect "powerless" statement there was changed in https://github.com/w3c/csswg-drafts/commit/96be9333a3b9b63f84d480b1f2df948975ee3ee6, discussion at https://github.com/w3c/csswg-drafts/issues/8609.

The reason Oklab/Lab are clamped appears to be that they're not great at predicting colors outside of SDR. The reason HSL became unbounded is because it roundtrips fine to sRGB.

I believe Lab and Oklab in theory also roundtrip fine, they're just matrix multiplies and cubes, though they get into floating point issues sooner.

PR #154, opened 2025-04-19
Clamp negative percentages in HSL

LaurenzV:
From my understanding of the [spec](https://www.w3.org/TR/css-color-4/#the-hsl-notation), both, S and L need to be clamped to a valid range:

image

After making this change, we get the same result as the current conversion code in resvg (I'm attempting to integrate this crate into `svgtypes`).

### DIFF
--- a/color/src/parse.rs
+++ b/color/src/parse.rs
@@ -429,7 +429,7 @@ impl<'a> Parser<'a> {
         let mode = if comma { Mode::Legacy } else { Mode::Modern };
         let s = self.scaled_component(1., 1.)?.map(|x| x.max(0.));
         self.optional_comma(comma)?;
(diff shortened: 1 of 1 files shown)


tomcur (commented):
I believe the current specification says not to clamp lightness, though it did used to specify clamping.

The change was made in https://github.com/w3c/csswg-drafts/commit/44c37787dfeeb299bad9be0f2dd7290f0f76bb53, also see the discussion at https://github.com/w3c/csswg-drafts/issues/8444.

Also see this issue: https://github.com/linebender/color/issues/106.

LaurenzV:
Ah, I hadn't seen that issue. So I suppose the current behavior is correct? In that case I will close the PR, thanks!

PR #155, opened 2025-04-19
Implement From<AlphaColor> for dynamic color

LaurenzV:
Not sure if there is a reason why this doesn't already exists, but it's much nicer if you can do `BLACK.into()` instead of having to do `DynamicColor::from_alpha_color(BLACK)`.

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -355,7 +355,7 @@ impl ColorSpace for LinearSrgb {
 
 impl From<LinearSrgb> for ColorSpaceTag {
     fn from(_: LinearSrgb) -> Self {
-        ColorSpaceTag::LinearSrgb
+        Self::LinearSrgb
     }
 }
 
@@ -420,7 +420,7 @@ impl ColorSpace for Srgb {
 
 impl From<Srgb> for ColorSpaceTag {
     fn from(_: Srgb) -> Self {
-        ColorSpaceTag::Srgb
+        Self::Srgb
     }
 }
 
@@ -472,7 +472,7 @@ impl ColorSpace for DisplayP3 {
 
 impl From<DisplayP3> for ColorSpaceTag {
     fn from(_: DisplayP3) -> Self {
-        ColorSpaceTag::DisplayP3
+        Self::DisplayP3
     }
 }
 
@@ -552,7 +552,7 @@ impl ColorSpace for A98Rgb {
 
 impl From<A98Rgb> for ColorSpaceTag {
     fn from(_: A98Rgb) -> Self {
-        ColorSpaceTag::A98Rgb
+        Self::A98Rgb
     }
 }
 
@@ -627,7 +627,7 @@ impl ColorSpace for ProphotoRgb {
 
 impl From<ProphotoRgb> for ColorSpaceTag {
     fn from(_: ProphotoRgb) -> Self {
-        ColorSpaceTag::ProphotoRgb
+        Self::ProphotoRgb
     }
 }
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
From previous discussions with @nicoburns and @raphlinus , the objection here was that if the color space doesn't have a tag, then this is not a lossless conversion and so we didn't want to have a `From` impl:

> The "most correct" thing would be ∀ CS From<AlphaColor<CS>> for DynamicColor where CS::Tag.is_some()

But this does come often! (This is the third time, I think.)

LaurenzV:
I guess I can change the PR to add a comment with this reasoning instead?

nicoburns:
Perhaps we could represent it as ` TryFrom` implementation (which would also make it clear why `From` can't be implemented).

nicoburns:
We could also implement `From<ColorSpace> for ColorSpaceTag` for each color space that has a tag. Then define:

```rust
impl<CS> From<AlphaColor<CS>> for DynamicColor where ColorSpaceTag: From<CS>
```

tomcur (commented):
I agree with `From` ideally being lossless. I'm happy with `TryFrom`, failing if the space has no tag. Nico's suggestion of statically enumerating the possible conversions is also interesting.

waywardmonkeys (commented):
I like this more. I think we will end up with questions about always unliking stuff like this but that should be a follow up that is consistently applied across the repo.

color/src/colorspace.rs
```diff
@@ -353,6 +353,12 @@ impl ColorSpace for LinearSrgb {
     }
 }
 
+impl From<LinearSrgb> for ColorSpaceTag {
+    fn from(_: LinearSrgb) -> Self {
```
DJMcNab, line 357: In theory, this could be:
```rust
    fn from(LinearSrgb: LinearSrgb) -> Self {
```
which would make it robust against `LinearSrgb` gaining fields in the future (this would fail to compile in that case). However, I think that's sufficiently unlikely that we don't really need to worry about this.

That is, I don't think we should do this (but I wouldn't review against it if it were the state)

color/src/dynamic.rs
```diff
@@ -447,6 +447,18 @@ impl BitHash for DynamicColor {
     }
 }
 
+// Note that the conversion is only lossless for color spaces that have a tag,
+// this is why we have this additional trait bound. See also
+// https://github.com/linebender/color/pull/155 for more discussion.
```
DJMcNab, line 452: I think making this a doc comment would be fine

color/src/dynamic.rs
```diff
@@ -447,6 +447,18 @@ impl BitHash for DynamicColor {
...
+impl<CS: ColorSpace> From<AlphaColor<CS>> for DynamicColor
+where
+    ColorSpaceTag: From<CS>,
+{
+    fn from(value: AlphaColor<CS>) -> Self {
+        Self::from_alpha_color(value)
```
DJMcNab, line 458: Can we have `const { assert!(CS::TAG.is_some()) }` here? This is only a sanity check against mistaken impls of `From`; it's not about making this a potentially fallible impl which would resolve that at monomorphisation time.

DJMcNab (commented):
This makes sense to me. Not approving only because I'd like @tomcur to take a look.

tomcur (approved):
This looks good and is useful, thanks! One suggestion on the docs of the `From` impl.

waywardmonkeys (approved):
Looks good to me once Tom's doc edit is in!

LaurenzV:
Ah sorry that somehow slipped under my radar, will do it in a bit!

PR #156, opened 2025-04-22
Add absolute conversion matrices for ProPhoto RGB

tomcur:
This builds on https://github.com/linebender/color/pull/139, which added absolute color conversions (conversions without chromatic adaptation). The default method implementations added there use the chromatically-adapting methods and undo adaptation through additional matrix multiplies. For efficiency, we can do the absolute conversions using matrices that do not include adaptation in the first place.

This PR does that for ProPhoto RGB. More to follow.

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -552,6 +552,24 @@ impl ColorSpace for A98Rgb {
 #[derive(Clone, Copy, Debug)]
 pub struct ProphotoRgb;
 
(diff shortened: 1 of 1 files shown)


waywardmonkeys (approved):
Assuming the tests validate that the matrix values are correct, looks good to me.

tomcur:
> Assuming the tests validate that the matrix values are correct, looks good to me.

They do, but can't catch small errors.

Issue #134, opened 2025-01-18
Remove deprecated `From<Rgba8> for PremulColor<Srgb>`


tomcur:
A reminder.

PR #157, opened 2025-04-22
Remove deprecated `From<Rgba8> for PremulColor<Srgb>`

tomcur:
Users should use `From<PremulRgba8> for PremulColor<Srgb>` instead.

Resolves https://github.com/linebender/color/issues/134.


DJMcNab (approved):
Originally derepcated by #113 

Removing this is correct.

waywardmonkeys (approved):

PR #159, opened 2025-04-23
Improve `gradient` documentation

tomcur:
Explain usage before going into the motivation, explain parameters, link to spec, and add an example comparing this to regular interpolation (testing at the same time).

### DIFF
--- a/color/src/gradient.rs
+++ b/color/src/gradient.rs
@@ -27,7 +27,7 @@ pub struct GradientIter<CS: ColorSpace> {
 /// Generate a piecewise linear approximation to a gradient ramp.
 ///
 /// The target gradient ramp is the linear interpolation from `color0` to `color1` in the target
-/// color space specified by `interp_cs`. For efficiency, this returns an
+/// color space specified by `interp_cs`. For efficiency, this function returns an
 /// [iterator over color stops](GradientIter) in the `CS` color space, such that the gradient ramp
 /// created by linearly interpolating between those stops in the `CS` color space is equal within
 /// the specified `tolerance` to the target gradient ramp.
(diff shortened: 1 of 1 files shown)


color/src/gradient.rs
```diff
@@ -26,26 +26,84 @@ pub struct GradientIter<CS: ColorSpace> {
...
+/// color space specified by `interp_cs`. For efficiency, this returns an
+/// [iterator over color stops](GradientIter) in the `CS` color space, such that the gradient ramp
+/// created by linearly interpolating between those stops in the `CS` color space is equal within
+/// the specified `tolerance` to the target gradient ramp.
+///
+/// When the target interpolation color space is cylindrical, the hue can be interpolated in
+/// multiple ways. The [`direction`](`HueDirection`) parameter controls the way in which the hue is
+/// interpolated.
+///
+/// The given `tolerance` value specifies the maximum error in the approximation, in deltaEOK
```
waywardmonkeys, line 39: I think we should link to something to explain terms like "deltaEOK" ... perhaps https://www.w3.org/TR/css-color-4/#color-difference-OK ?

waywardmonkeys (approved):
Overall, this seems great.  A minor nit about linking to definition of "deltaEOK".

It might be that we should deal with deltaEOK linking separate from this PR though.

color/src/gradient.rs
tomcur: I agree. I've added an explanation and linked to that section in the spec. Perhaps some of this stuff should be expanded on elsewhere in the future: for now we just have `gradient` where this is used, but it's also relevant for tone mapping.

waywardmonkeys (approved):

PR #166, opened 2025-04-30
Faster rounding in `{AlphaColor,OpaqueColor,PremulColor}::to_rgba8`

### LINKED ISSUE #142: Conversion to RGBA8 is slow on x86 targets
See https://github.com/linebender/vello/pull/830#discussion_r1982017666 for more information.

tomcur:
Resolves #142. See also the discussion at https://github.com/linebender/vello/pull/830#discussion_r1982017666.

On x86, `f32::round` is slow as it calls out to the IEEE754-adhering libc function `roundf`. See, e.g., the branchy [glibc `roundf`](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/ieee754/flt-32/s_roundf.c;h=eb714e975e48f737c140a1dd666135759ac0ffe3;hb=HEAD) and [musl `roundf`](https://git.etalabs.net/cgit/musl/tree/src/math/roundf.c?h=c47ad25ea3b484e10326f933e927c0bc8cded3da) implementations. On AArch64, the compiler can make use of hardware instructions instead.

The knowledge that the output range is `0-255`, allows writing an almost-correct rounding implementation that doesn't require calling out to libc or branching. This is fast on both x86 and AArch64.

Codegen:
x86: https://godbolt.org/z/PaEP43fGP
AArch64: https://godbolt.org/z/vd9Wcjz6q

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -1029,7 +1029,7 @@ mod tests {
             v.round() as u8
         }
 
-        // Check the rounding behavior of als floating point values within (and near) the range
+        // Check the rounding behavior of all floating point values within (and near) the range
         // 0-255.
         let mut failures = alloc::vec![];
         let mut v = -1_f32;
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -1019,4 +1019,28 @@ mod tests {
 
         assert_eq!(&failures, &[0.49999997]);
     }
+
+    /// If you don't trust the test above, run this one by removing `#[ignore]`.
+    #[test]
+    #[ignore]
+    fn fast_round_full() {
+        #[expect(clippy::cast_possible_truncation, reason = "deliberate quantization")]
+        fn real_round_to_u8(v: f32) -> u8 {
+            v.round() as u8
+        }
+
+        // Check the rounding behavior at integer and half integer values within (and near) the
+        // range 0-255, as well as one ULP up and down from those values.
+        let mut failures = alloc::vec![];
+        let mut v = -1_f32;
+
+        while v <= 256. {
+            if real_round_to_u8(v) != fast_round_to_u8(v) {
+                failures.push(v);
+            }
+            v = v.next_up();
+        }
+
+        assert_eq!(&failures, &[0.49999997]);
+    }
 }
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -1029,8 +1029,8 @@ mod tests {
             v.round() as u8
         }
 
-        // Check the rounding behavior at integer and half integer values within (and near) the
-        // range 0-255, as well as one ULP up and down from those values.
+        // Check the rounding behavior of als floating point values within (and near) the range
+        // 0-255.
         let mut failures = alloc::vec![];
         let mut v = -1_f32;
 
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
FYI @LaurenzV

LaurenzV (approved):
Thanks!

color/src/color.rs
```diff
@@ -630,20 +647,40 @@ impl<CS: ColorSpace> PremulColor<CS> {
...
 }
 
+/// Fast rounding of `f32` to integer `u8`, rounding ties up.
+///
+/// Targeting x86, `f32::round` calls out to libc `roundf`. Even if that call were inlined, it is
+/// branchy, which would make it relatively slow. The following is faster, and on the range `0-255`
+/// almost correct*. AArch64 has dedicated rounding instructions so does not need this
+/// optimization, but the following is still fast.
+///
+/// * The only input where the output differs from `a.round() as u8` is `0.49999997`.
```
DJMcNab, line 676: Would it be reasonable to commit a test here? 

My suggested test would be a surgical once, which goes up in increments of a half and asserts that a few floats either side have the correct value (i.e. the same as `f32::round`, and if that test fails, `panic` unless the value is exactly `0.49999997`). That should be reasonably quick to run, and means that this claim doesn't need to be taken on trust.

DJMcNab (commented):
One thought about a test here, which needn't block. Can't comment on whether we want this, other than based on a recollection from last week's meeting

color/src/color.rs
```diff
@@ -939,4 +981,42 @@ mod tests {
...
+
+            if real_round_to_u8(down) != fast_round_to_u8(down) {
+                failures.push(down);
+            }
+            if real_round_to_u8(v) != fast_round_to_u8(v) {
+                failures.push(v);
+            }
+            if real_round_to_u8(up) != fast_round_to_u8(up) {
+                failures.push(up);
+            }
```
DJMcNab, line 1015: Sorry to be picky, but I might prefer us to go two ULPs in each direction (i.e. to validate that `0.49999994` has the correct behaviour). I think that the main testing block can be made into a closure, so it would look a bit like:

```
let validate_rounding = |val: f32| if real_round_to_u8(val) != fast_round_to_u8(val) {failures.push(val)};

validate_rounding(v.next_down().next_down());
validate_rounding(v.next_down());
validate_rounding(v);
validate_rounding(v.next_up());
validate_rounding(v.next_up().next_up());
```

color/src/color.rs
```diff
@@ -939,4 +981,66 @@ mod tests {
...
+
+            v += 0.5;
+        }
+
+        assert_eq!(&failures, &[0.49999997]);
+    }
+
+    /// If you don't trust the test above, run this one by removing `#[ignore]`.
+    #[test]
+    #[ignore]
```
DJMcNab, line 1025:
```suggestion
    /// A more thorough test than the one above, which tests only the values which are likely to fail.
    /// This test runs through approximately 200 million floats, so can be somewhat slow (seconds 
    /// rather than milliseconds). To run this test, use the `--ignored` flag.
    #[test]
    #[ignore = "Takes too long to execute."]
```

PR #171, opened 2025-05-01
Specify XYZ-D65<->sRGB conversion as exact rationals

tomcur:


DJMcNab (approved):
I have confirmed that these are "visually" the same numbers as before (i.e. I've scanned the first ~3 significant digits) of each.

This seems consistent with other changes we've done.

PR #175, opened 2025-05-19
Fix `core_float_math` nightly compilation.

ajakubowicz-canva:
### Context

This PR fixes an error in CI when using nightly toolchain:

```
error[E0658]: use of unstable library feature `core_float_math`
```

This error can be seen in `vello` here:
  - E.g. https://github.com/linebender/vello/actions/runs/15104337336/job/42450338042

### Fix

Fix was entirely proposed by @DJMcNab here: [#vello > cargo doc CI failing with rust nightly @ 💬](https://xi.zulipchat.com/#narrow/channel/197075-vello/topic/cargo.20doc.20CI.20failing.20with.20rust.20nightly/near/519001092)

### Test plan

CI

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -16,7 +16,7 @@ use crate::floatfuncs::FloatFuncs;
 /// Use `F32::cbrt` to call a method on the `f32` type, rather than accessing a function from
 /// `core::f32` with the same name. Currently `core::f32` implements unstable `core_float_math`
 /// features that break CI.
-/// Issue: https://github.com/rust-lang/rust/issues/137578
+/// Issue: <https://github.com/rust-lang/rust/issues/137578>
 type F32 = f32;
 
 /// The main trait for color spaces.
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -13,7 +13,7 @@ use crate::floatfuncs::FloatFuncs;
 
 /// Type alias to disambiguate between the `f32` primitive type and `core::f32` module.
 ///
-/// Use `F32::cbrt` to call a method on the `f32`` type, rather than accessing a function from
+/// Use `F32::cbrt` to call a method on the `f32` type, rather than accessing a function from
 /// `core::f32` with the same name. Currently `core::f32` implements unstable `core_float_math`
 /// features that break CI.
 /// Issue: https://github.com/rust-lang/rust/issues/137578
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -11,6 +11,14 @@ use crate::{matvecmul, tag::ColorSpaceTag, Chromaticity};
 #[cfg(all(not(feature = "std"), not(test)))]
 use crate::floatfuncs::FloatFuncs;
 
+/// Type alias to disambiguate between the `f32` primitive type and `core::f32` module.
+///
+/// Use `F32::cbrt` to call a method on the `f32`` type, rather than accessing a function from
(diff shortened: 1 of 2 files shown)


color/src/colorspace.rs
```diff
@@ -337,7 +345,7 @@ impl ColorSpace for LinearSrgb {
     }
 
     fn scale_chroma(src: [f32; 3], scale: f32) -> [f32; 3] {
-        let lms = matvecmul(&OKLAB_SRGB_TO_LMS, src).map(f32::cbrt);
+        let lms = matvecmul(&OKLAB_SRGB_TO_LMS, src).map(F32::cbrt);
```
ajakubowicz-canva, line 348: On nightly, `core::f32::cbrt` is unstable and requires the `core_float_math` nightly feature. Hence by disambiguating to the `f32` type, we ensure that we use `cbrt` defined on the type (implemented via `FloatFuncs` trait), instead of the unstable `core:f32:cbrt` method.

waywardmonkeys (approved):

waywardmonkeys:
We should probably do a patch release this week with this.

xStrom:
I think that removing the `core::f32` import is a better solution.

```diff
diff --git a/color/src/colorspace.rs b/color/src/colorspace.rs
index ddb6a60..24dc127 100644
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -4,7 +4,7 @@
 #[cfg(feature = "std")]
 extern crate std;

-use core::{any::TypeId, f32};
+use core::{any::TypeId, f32::consts::PI};

 use crate::{matvecmul, tag::ColorSpaceTag, Chromaticity};

@@ -1253,7 +1253,7 @@ impl From<Oklab> for ColorSpaceTag {

 /// Rectangular to cylindrical conversion.
 fn lab_to_lch([l, a, b]: [f32; 3]) -> [f32; 3] {
-    let mut h = b.atan2(a) * (180. / f32::consts::PI);
+    let mut h = b.atan2(a) * (180. / PI);
     if h < 0.0 {
         h += 360.0;
     }
@@ -1263,7 +1263,7 @@ fn lab_to_lch([l, a, b]: [f32; 3]) -> [f32; 3] {

 /// Cylindrical to rectangular conversion.
 fn lch_to_lab([l, c, h]: [f32; 3]) -> [f32; 3] {
-    let (sin, cos) = (h * (f32::consts::PI / 180.)).sin_cos();
+    let (sin, cos) = (h * (PI / 180.)).sin_cos();
     let a = c * cos;
     let b = c * sin;
     [l, a, b]
```

ajakubowicz-canva:
@xStrom Agree – updating this PR.

waywardmonkeys:
Having both is best to avoid someone adding back a use of f32 and re-breaking it

xStrom:
They can't add it back that easily, it won't pass CI.

waywardmonkeys:
I mean both the F32 change and the removal that Kaur suggested.

xStrom (approved):
Go ahead and merge this. 🎉

xStrom:
I guess it's possible yes, if we suddenly become very conservative with our MSRV here so it doesn't reach this issue by the time of stabilization. As we don't test "middle" Rust versions.

However, the issue won't be unique to `colorspace.rs` or even just `color`. I think we'll need to be aware of this footgun in a very broad sense.

xStrom:
Added a changelog entry to this PR branch.

waywardmonkeys:
LGTM

PR #182, opened 2025-06-26
Work around CI failing due to float precision randomization

tomcur:
Miri now randomizes floating point precision of some operations that do not have guaranteed precision. That's causing CI to fail.

This works around the issue by using a looser bound when targeting Miri. That's a bit messy, but we'd probably like to notice when our normal targets suddenly fail to reach tight precision bounds.

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -1716,7 +1716,17 @@ mod tests {
     fn roundtrip() {
         fn test_roundtrips<Source: ColorSpace, Dest: ColorSpace>(colors: &[[f32; 3]]) {
             /// A tight bound on relative numerical precision.
+            #[cfg(not(miri))]
             const RELATIVE_EPSILON: f32 = f32::EPSILON * 16.;
+
+            /// Miri uses precision randomization of floating point operations that do not have a
+            /// guaranteed precision.
+            ///
+            /// Use a lower precision bound on Miri specifically. This is a bit messy, but on our
+            /// target platforms, we'd still like to notice it we don't reach a tight precision
+            /// bound anymore.
+            #[cfg(miri)]
+            const RELATIVE_EPSILON: f32 = f32::EPSILON * 1600.;
 
             for color in colors {
                 let intermediate = Source::convert::<Dest>(*color);


color/src/colorspace.rs
```diff
@@ -1716,8 +1716,18 @@ mod tests {
...
             const RELATIVE_EPSILON: f32 = f32::EPSILON * 16.;
 
+            /// Miri uses precision randomization of floating point operations that do not have a
+            /// guaranteed precision.
+            ///
+            /// Use a lower precision bound on Miri specifically. This is a bit messy, but on our
+            /// target platforms, we'd still like to notice it we don't reach a tight precision
+            /// bound anymore.
+            #[cfg(miri)]
+            const RELATIVE_EPSILON: f32 = f32::EPSILON * 1600.;
```
tomcur, line 1729: Bumps the bound by two orders of magnitude. I didn't base that on anything in specific – one order of magnitude also works, but perhaps that's flaky.

tomcur:
We discussed in Office Hours to merge this optimistically, and revisit next Office Hours when more people are available.

color/src/colorspace.rs
```diff
@@ -1716,8 +1716,18 @@ mod tests {
     fn roundtrip() {
         fn test_roundtrips<Source: ColorSpace, Dest: ColorSpace>(colors: &[[f32; 3]]) {
             /// A tight bound on relative numerical precision.
+            #[cfg(not(miri))]
             const RELATIVE_EPSILON: f32 = f32::EPSILON * 16.;
 
+            /// Miri uses precision randomization of floating point operations that do not have a
+            /// guaranteed precision.
+            ///
+            /// Use a lower precision bound on Miri specifically. This is a bit messy, but on our
```
DJMcNab, line 1725: I think that "lower" is a potentially confusing term here (given that the absolute value is higher, as it were). Looser is the term I'd normally expect to see here, but it's likely that there's some subtlety here I'm not familiar with.

DJMcNab (approved):
Approving based on discussion in office hours.

Neither comment is blocking; I don't think this is worth spending too much mental effort on

PR #185, opened 2025-07-30
 Support interpolating in straight alpha space by accepting `AlphaInterpolationSpace` argument

### LINKED ISSUE #184: Add support for interpolating gradients in non-premultiplied space.
See https://github.com/linebender/vello/issues/1129.

sagudev:
Implementation is pretty simple, we premul step into before interpolation and only do it if actually needed (`AlphaInterpolationSpace::Premultiplied`). I also added unpremultiplied test.

I used dynamic (or alpha) color everywhere as even [CSS that does interpolation in premultiplied alpha converts it back to unpremultiplied](https://www.w3.org/TR/css-color-4/#interpolation):

> changing the color components to [premultiplied](https://www.w3.org/TR/css-color-4/#premultiplied) form
> linearly interpolating each component of the computed value of the color separately
> undoing [premultiplication](https://www.w3.org/TR/css-color-4/#premultiplied)

Fixes #184

This is breaking change.

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -107,6 +107,56 @@ pub enum HueDirection {
     // NOTICE: If a new value is added, be sure to modify `MAX_VALUE` in the bytemuck impl.
 }
 
+/// Defines how color channels should be handled when interpolating
+/// between transparent colors.
+#[derive(Clone, Copy, Default, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[repr(u8)]
+#[expect(
+    clippy::exhaustive_enums,
+    reason = "There are only two ways to do interpolation."
+)]
+pub enum AlphaInterpolationSpace {
+    /// Colors are interpolated with their color channels premultiplied by the alpha
+    /// channel. This is almost always what you want.
+    ///
+    /// Used when interpolating colors in the premultiplied alpha space, which allows
+    /// for correct interpolation when colors are transparent. This matches behavior
+    /// described in [CSS Color Module Level 4 § 12.3].
+    ///
+    /// Following the convention of CSS Color Module Level 4, in cylindrical color
+    /// spaces the hue channel is not premultiplied. If it were, interpolation would
+    /// give undesirable results. See also [`PremulColor`].
+    ///
+    /// [CSS Color Module Level 4 § 12.3]: https://drafts.csswg.org/css-color/#interpolation-alpha
+    #[default]
+    Premultiplied = 0,
+    /// Colors are interpolated without premultiplying their color channels by the alpha channel.
+    ///
+    /// This causes color information to leak out of transparent colors. For example, when
+    /// interpolating from a fully transparent red to a fully opaque blue in sRGB, this
+    /// method will go through an intermediate purple.
+    ///
+    /// Used when interpolating colors in the unpremultiplied (straight) alpha space.
+    /// This matches behavior of gradients in the HTML `canvas` element.
+    /// See [The 2D rendering context § Fill and stroke styles].
+    ///
+    /// [The 2D rendering context § Fill and stroke styles]: https://html.spec.whatwg.org/multipage/#interpolation
+    Unpremultiplied = 1,
+}
+
+impl AlphaInterpolationSpace {
+    /// Returns `true` if the alpha mode is [`AlphaInterpolationSpace::Premultiplied`].
+    pub const fn is_premultiplied(&self) -> bool {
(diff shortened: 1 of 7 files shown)


color/src/color.rs
```diff
@@ -107,6 +107,42 @@ pub enum HueDirection {
...
 
+/// Defines how to handle alpha during interpolation.
+#[derive(Clone, Copy, Default, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[repr(u8)]
+#[expect(
+    clippy::exhaustive_enums,
+    reason = "There are only two ways to do interpolation."
+)]
+pub enum AlphaInterpolationSpace {
```
sagudev, line 118: Should we name this `AlphaType` so that we can use it in other places too?

DJMcNab:
(This will absolutely need a changelog entry)

color/src/color.rs
nicoburns: I'm thinking that we may well want to use this in the `Image` type (along with the existing `ImageFormat` specifier).

color/src/color.rs
```diff
@@ -107,6 +107,46 @@ pub enum HueDirection {
...
 
+/// Defines possible handling of alpha channel
+#[derive(Clone, Copy, Default, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[repr(u8)]
+#[expect(
+    clippy::exhaustive_enums,
+    reason = "There are only two ways to do interpolation."
+)]
+pub enum AlphaMode {
```
sagudev, line 118: Writing docs and naming was hard here. I also kept interpolation space explanation here.

tomcur (commented):
Thanks, I think it makes sense to add this, especially given the HTML `canvas` documentation you linked. I haven't had the time yet to dive into the math, but I've suggested a few documentation tweaks inline.

color/src/color.rs
```diff
@@ -107,6 +107,46 @@ pub enum HueDirection {
...
+    #[default]
+    Premultiplied = 0,
+    /// Straight (unpremultiplied) alpha.
+    /// Color values are stored independently of the alpha channel.
+    ///
+    /// Used when interpolating colors in straight alpha space.
+    /// This matches behavior described in [The 2D rendering context § Fill and stroke styles].
+    ///
+    /// [The 2D rendering context § Fill and stroke styles]: https://html.spec.whatwg.org/multipage/#interpolation
+    Unpremultiplied = 1,
```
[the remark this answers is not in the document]
tomcur: (Though also see https://github.com/linebender/color/pull/185#discussion_r2259623445.)

color/src/color.rs
sagudev: Actually we probably want to use this AlphaMode in other cases too, like https://github.com/linebender/color/pull/185#discussion_r2258468178.

sagudev:
`AlphaMode` is intended to be more than just a way to provide interpolation alpha space, see generalization I've done in https://github.com/linebender/color/pull/185/commits/26e5c0bd023c6c03b74bb140db2d930cc25b217e. We can also use it for images as suggested here: https://github.com/linebender/color/pull/185#discussion_r2258468178

tomcur:
> `AlphaMode` is intended to be more than just a way to provide interpolation alpha space, see generalization I've done in [26e5c0b](https://github.com/linebender/color/commit/26e5c0bd023c6c03b74bb140db2d930cc25b217e). We can also use it for images as suggested here: [#185 (comment)](https://github.com/linebender/color/pull/185#discussion_r2258468178)

Thanks, I've responded to your other comment about that here: [here ](https://github.com/linebender/color/pull/185#discussion_r2259804256).

color/src/color.rs
tomcur: This is a good reason for the change of the original name of `AlphaInterpolationSpace` to `AlphaMode`, but on the other hand, it feels as if it's now slightly conflating two things: namely how color channels are stored (which should not impact end results), and how color channels are treated during interpolation (which definitely results in observable behavior). It may not matter too much, though.

nicoburns: Yes, good point.

sagudev: I will move back to `AlphaInterpolationSpace`

DJMcNab (commented):
Broadly, this is looking reasonable, but I do have some concerns; I don't want this to negatively impact other users of Color, and I agree that we should strongly steer people to use Premultiplied.

(I'll leave the bulk of the review to Tom)

PR #186, opened 2025-08-03
to_degrees + to_radians instead of manual conversion

ProgramCrafter:
More readable, and should Rust `core` ever choose a faster/more precise algorithm it will be available right away.


xStrom:
Need to remove the `PI` import then to make Clippy happy.

PR #187, opened 2025-08-07
Test interpolation of transparent colors

tomcur:
This adds two tests interpolating from a fully transparent to a fully opaque color, in a rectangular space and in a cylindrical space.

This is motivated by https://github.com/linebender/color/pull/185, testing current behavior stays correct.


raphlinus (approved):
This is fine as soon as the .abs() is added on the asserts.

PR #188, opened 2025-08-07
More `to_degrees`

tomcur:
Follow-up to https://github.com/linebender/color/pull/186.

### DIFF
--- a/color/src/parse.rs
+++ b/color/src/parse.rs
@@ -317,7 +317,7 @@ impl<'a> Parser<'a> {
                 let dim_lc = make_lowercase(dim, &mut buf);
                 let scale = match dim_lc {
                     "deg" => 1.0,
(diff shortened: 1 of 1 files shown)


color/src/parse.rs
```diff
@@ -317,7 +317,7 @@ impl<'a> Parser<'a> {
                 let dim_lc = make_lowercase(dim, &mut buf);
                 let scale = match dim_lc {
                     "deg" => 1.0,
-                    "rad" => 180.0 / f64::consts::PI,
+                    "rad" => 1_f64.to_degrees(),
```
DJMcNab, line 320: How about:
```suggestion
                    "rad" => const { 1_f64.to_degrees(), }
```

tomcur: Good idea, will do tomorrow. I do hope it's redundant in practice!

PR #190, opened 2025-08-09
Better document `AlphaColor` vs `PremulColor` premultiplication

### LINKED ISSUE #189: Document if `AlphaColor` assumed to be pre-multiplied or not
See subject. :grin:

tomcur:
Fixes #189.

### DIFF
--- a/color/src/color.rs
+++ b/color/src/color.rs
@@ -40,6 +40,10 @@ pub struct OpaqueColor<CS> {
 ///
 /// A color in a color space known at compile time, with an alpha channel.
 ///
+/// The color channels are "straight," i.e., they are not premultiplied by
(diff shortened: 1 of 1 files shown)


tomcur:
> I guess there could be an `AlphaColorPremul` but that sounds weird. One would possibly expect `ColorAlphaPremul` which then makes `AlphaColor` stand out.

It is of course possible I'm missing something, but have you seen [`PremulColor`](https://docs.rs/color/latest/color/struct.PremulColor.html)?

https://github.com/linebender/color/blob/860c71795d9afc5832f8fe2ed5c0aeb2e8b3dc99/color/src/color.rs#L61-L90

PR #191, opened 2025-08-20
Support interpolating in unpremultiplied (straight) alpha space

### LINKED ISSUE #184: Add support for interpolating gradients in non-premultiplied space.

sagudev:
Alternative implementiran of #185 that is NOT a breaking change. Functional wise it's the same (although there is no generalization of gradient iter code).

Fixes #184

### DIFF
--- a/color/src/dynamic.rs
+++ b/color/src/dynamic.rs
@@ -356,20 +356,20 @@ impl DynamicColor {
     }
 
     /// Interpolate two colors.
-/// Interpolate two colors without alpha premultiplication.
-///
-/// Similar to [`interpolate`], but colors are interpolated without premultiplying their color
-/// channels by the alpha channel. This is almost never what you want.
-///
-/// This causes color information to leak out of transparent colors. For example, when
-/// interpolating from a fully transparent red to a fully opaque blue in sRGB, this
-/// method will go through an intermediate purple.
-///
-/// This matches behavior of gradients in the HTML `canvas` element.
-/// See [The 2D rendering context § Fill and stroke styles][HTML 2D Canvas] of the
-/// HTML 2D Canvas specification.
-///
-/// [HTML 2D Canvas]: https://html.spec.whatwg.org/multipage/#interpolation
+    /// Interpolate two colors without alpha premultiplication.
+    ///
+    /// Similar to [`interpolate`], but colors are interpolated without premultiplying their color
+    /// channels by the alpha channel. This is almost never what you want.
+    ///
+    /// This causes color information to leak out of transparent colors. For example, when
+    /// interpolating from a fully transparent red to a fully opaque blue in sRGB, this
+    /// method will go through an intermediate purple.
+    ///
+    /// This matches behavior of gradients in the HTML `canvas` element.
+    /// See [The 2D rendering context § Fill and stroke styles][HTML 2D Canvas] of the
+    /// HTML 2D Canvas specification.
+    ///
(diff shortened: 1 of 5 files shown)


color/src/dynamic.rs
```diff
@@ -347,6 +351,43 @@ impl DynamicColor {
...
+    /// let mid = interp.eval(0.5);
+    /// assert_eq!(mid.cs, ColorSpaceTag::Hsl);
+    /// assert!((mid.components[0] - 60.).abs() < 0.01);
+    /// ```
+    pub fn interpolate_with_alpha(
+        self,
+        other: Self,
+        cs: ColorSpaceTag,
+        direction: HueDirection,
+        interpolation_alpha_space: InterpolationAlphaSpace,
```
tomcur, line 390: Perhaps like `unpremultiplied_gradient` this should be `unpremultiplied_interpolate` and not take the enum. That prevents having to make the `InterpolationAlphaSpace` public.

(Though if we're renaming, `interpolate_unpremultiplied` sounds nicer, so perhaps it should also be `gradient_unpremultiplied`.)

tomcur (commented):
I think I prefer this over #184! We discussed this in today's office hours, and having these separate functions as special-case "quirks" seems nice, as users don't need to spend the mental overhead each time they're using the normal functions.

tomcur (commented):
Thanks! This looks good. I've placed some comments inline, especially about how we communicate this.

The following are two non-blocking comments.

Like the `UnpremultipliedGradientIter` you have introduced, I think it may be worthwhile to explore introducing an `UnpremultipliedInterpolator`, to lift some branching to the type level.

It perhaps may be worthwhile to also explore deduplicating the code between the two gradient iters; perhaps by extracting everything to some function that does the bulk of that work. The painful thing is that it currently would have to be mostly type-erased, or do something hacky, as we have two separate types (`AlphaColor` and `PremulColor`) that we can't currently be generic over... this is related to the discussion in https://github.com/linebender/color/pull/190, and specifically your comment there.

color/src/gradient.rs
```diff
@@ -181,3 +184,158 @@ impl<CS: ColorSpace> Iterator for GradientIter<CS> {
...
+/// channels by the alpha channel. This is almost never what you want.
+///
+/// This causes color information to leak out of transparent colors. For example, when
+/// interpolating from a fully transparent red to a fully opaque blue in sRGB, this
+/// method will go through an intermediate purple.
+///
+/// This matches behavior of gradients in the HTML `canvas` element.
+/// See [The 2D rendering context § Fill and stroke styles][HTML 2D Canvas] of the
+/// HTML 2D Canvas specification.
+/// [HTML 2D Canvas]: <https://html.spec.whatwg.org/multipage/#interpolation>
```
tomcur, line 226: A newline is required to get the link to render here.

```suggestion
/// This matches behavior of gradients in the HTML `canvas` element.
/// See [The 2D rendering context § Fill and stroke styles][HTML 2D Canvas] of the
/// HTML 2D Canvas specification.
///
/// [HTML 2D Canvas]: <https://html.spec.whatwg.org/multipage/#interpolation>
```

tomcur (approved):
This is great, thank you!

This can be merged once the remaining inline comments have been fixed.

A few documentation lines got duplicated when the suggestions were applied, I've listed the ones I found inline. I think I caught them all.

sagudev:
I rechecked and no lines are duplicated anymore so landing this.

PR #194, opened 2025-09-10
Bump dependencies for release

tomcur:


DJMcNab (approved):
This does help make sure that our MSRV is still accurate, I guess? But otherwise, it has no impact on the results of using this package, right?

PR #195, opened 2025-09-10
Use stronger wording in unpremultiplied interpolation changelog entry

tomcur:


CHANGELOG.md
```diff
@@ -18,7 +18,10 @@ This release has an [MSRV][] of 1.82.
 ### Added
 
 * Add `interpolate_unpremultiplied` and `gradient_unpremultiplied` for interpolating in unpremultiplied (straight) alpha space.
-  This allows using Color to implement web rendering features which specify such gradient rendering. ([#185][] by [@sagudev][])
+
+  While such interpolation will often give perceptually undesired results, this allows using Color to implement rendering features where such interpolation is specified or desired, like in the [HTML Canvas element][]. ([#185][] by [@sagudev][])
+
+[HTML Canvas element]: https://html.spec.whatwg.org/commit-snapshots/a93c6fa9fa95e31f1caa05f2f8abc650669df7c3/#interpolation
```
tomcur, line 24: Our links are generally at the bottom of the changelog. Do we want that for this, too? To the extreme it could get somewhat confusing.

DJMcNab: I think per-section links here makes sense.

I'd probably put it at the bottom of "Unreleased", rather than at the bottom of "Added", but that isn't based on any strong principles.

CHANGELOG.md
```diff
@@ -18,7 +18,10 @@ This release has an [MSRV][] of 1.82.
 ### Added
 
 * Add `interpolate_unpremultiplied` and `gradient_unpremultiplied` for interpolating in unpremultiplied (straight) alpha space.
-  This allows using Color to implement web rendering features which specify such gradient rendering. ([#185][] by [@sagudev][])
+
+  While such interpolation will often give perceptually undesired results, this allows using Color to implement rendering features where such interpolation is specified or desired, like in the [HTML Canvas element][]. ([#185][] by [@sagudev][])
```
DJMcNab, line 22:
```suggestion
  While such interpolation will often give perceptually undesired results, this allows using Color to implement rendering features where such interpolation is specified, like in the [HTML Canvas element][]. ([#185][] by [@sagudev][])
```

To make it even stronger. This is non-blocking; I'm not even convinced it's an improvement.

DJMcNab (approved):
Looks good, thanks!

CHANGELOG.md
tomcur: > I'm not even convinced it's an improvement.

Fewer words!

PR #207, opened 2026-02-04
Improve hue handling in `white_components` test

tomcur:
Hue is "powerless" when the color is gray, and depending on the exact math can jitter quite a bit. In this test the color is white by definition. By zeroing hue out and making the epsilon dependent on the other components' magnitudes, we can have a tight relative bound for this test.

The main motivator of this is to attempt re-enabling Miri's float randomization (it passes locally with this change). See https://github.com/linebender/color/pull/196.

### DIFF
--- a/color/src/colorspace.rs
+++ b/color/src/colorspace.rs
@@ -1791,16 +1791,29 @@ mod tests {
     #[test]
     fn white_components() {
         fn check_white<CS: ColorSpace>() {
-            assert!(almost_equal::<Srgb>(
-                Srgb::WHITE_COMPONENTS,
-                CS::convert::<Srgb>(CS::WHITE_COMPONENTS),
-                1e-4,
-            ));
-            assert!(almost_equal::<CS>(
-                CS::WHITE_COMPONENTS,
-                Srgb::convert::<CS>(Srgb::WHITE_COMPONENTS),
-                1e-4,
-            ));
+            assert!(
+                almost_equal::<Srgb>(
+                    Srgb::WHITE_COMPONENTS,
+                    CS::convert::<Srgb>(CS::WHITE_COMPONENTS),
+                    1e-5 * magnitude(Srgb::WHITE_COMPONENTS),
+                ),
+                "Failed for {:?}",
+                CS::TAG
+            );
+            // For cylindrical color spaces, hue is usually unstable when the color is gray in
+            // general, or white specifically as here, so zero out hue for the purposes of this
+            // test. Also see `DynamicColor::powerless_to_missing`.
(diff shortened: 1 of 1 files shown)


color/src/colorspace.rs
```diff
@@ -1791,16 +1791,29 @@ mod tests {
...
+                "Failed for {:?}",
+                CS::TAG
+            );
+            // For cylindrical color spaces, hue is usually unstable when the color is gray in
+            // general, or white specifically as here, so zero out hue for the purposes of this
+            // test. Also see `DynamicColor::powerless_to_missing`.
+            let mut expected = CS::WHITE_COMPONENTS;
+            let mut actual = Srgb::convert::<CS>(Srgb::WHITE_COMPONENTS);
+            if let Some(hue) = CS::LAYOUT.hue_channel() {
+                expected[hue] = 0.0;
```
DJMcNab, line 1809: Is this line actually needed? Should the CS::WHITE_COMPONENTS have a zero hue anyway?

color/src/colorspace.rs
```diff
@@ -1791,16 +1791,29 @@ mod tests {
...
-            assert!(almost_equal::<CS>(
-                CS::WHITE_COMPONENTS,
-                Srgb::convert::<CS>(Srgb::WHITE_COMPONENTS),
-                1e-4,
-            ));
+            assert!(
+                almost_equal::<Srgb>(
+                    Srgb::WHITE_COMPONENTS,
+                    CS::convert::<Srgb>(CS::WHITE_COMPONENTS),
+                    1e-5 * magnitude(Srgb::WHITE_COMPONENTS),
```
DJMcNab, line 1798: Just to clarify for myself, is this a tighter or looser bound?

DJMcNab (approved):
This seems fairly reasonable. It could be worth explaining the fact that we do this because of float randomisation in a comment, but that isn't blocking.

color/src/colorspace.rs
tomcur: Currently `Oklch` has a hue of 90° for white... but maybe that should be 0°. That's for a separate PR.

https://github.com/linebender/color/blob/32313f4881735407936cb05b0ab6e8865f1f3819/color/src/colorspace.rs#L1290-L1295

color/src/colorspace.rs
tomcur: This is tighter than the `1e-4` epsilon in `Srgb`-space before (`magnitude` is just `1` here), but the check we dropped in the generic `CS`-space was sometimes quite a bit looser *or* tighter than this (though it was flaky as the bound there was not well-defined).

PR #210, opened 2026-04-22
Set `Oklch::WHITE_COMPONENTS` to 0° hue

tomcur:
It's currently 90° on `main`, which is harmless as with 0 chroma it is powerless, but 0° is probably the more expected default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,11 @@ This release has an [MSRV][] of 1.82.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
I think we're reasonably happy to land this. Just to double check, there hasn't been any past discussion of this when it was added.

waywardmonkeys:
Tom and I talked about this some as it was me that had added these before. What we'd done at the time was to generate the `WHITE_COMPONENTS` for each colorspace by converting the white to that color space and printing out the values.

Since the hue is powerless in this case, it doesn't really matter what it is.

CHANGELOG.md
DJMcNab, line 1: Conclusion from today's office hours if that we don't want to mark either of these as breaking.
There's an argument for having a one-sentence comment about letting us know if #202 is an issue and we'll re-evaluate.

DJMcNab (approved):

PR #217, opened 2026-05-06
Add const `AlphaColor<sRGB>::from_hex` and `OpaqueColor<sRGB>::from_hex`

DJMcNab:
This is a rework of #212.

Many thanks to @l0uisgrange for kicking this off. I look forward to seeing your review.

I used Claude (Opus 4.6) to review (but not write) the docs entry, and to write a first version of the tests.


color/src/lib.rs
```diff
@@ -223,3 +345,79 @@ impl PremulColor<Srgb> {
 fn ensure_libm_dependency_used() -> f32 {
     libm::sqrtf(4_f32)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const ALPHA_FROM_HEX_IS_CONST: AlphaColor<Srgb> = AlphaColor::from_hex("#8a2be2");
+    #[test]
```
tomcur, line 335:
```suggestion
    const ALPHA_FROM_HEX_IS_CONST: AlphaColor<Srgb> = AlphaColor::from_hex("#8a2be2");
    
    #[test]
```

tomcur (approved):
This is useful, thanks!

Not blocking, but I think having an `Rgba8::try_from_hex` (and panicking counterpart) would be nicely symmetrical, where the `AlphaColor<Srgb>` method internally forwards to that and uses `AlphaColor<Srgb>::from_rgba8`. The parsing function `color_from_4bit_hex` effectively does something like that already.

PR #218, opened 2026-05-06
Bump MSRV to 1.86

DJMcNab:
This was released in April 2025, which is now more than a year ago. The reason I want this now is #217, but it
does also have user-facing improvements to constness.

I used Claude (Opus 4.6) to create the changelog entry.

### DIFF
--- a/color/src/floatfuncs.rs
+++ b/color/src/floatfuncs.rs
@@ -42,7 +42,6 @@ define_float_funcs! {
     fn hypot(self, other: Self) -> Self => hypotf;
     // Note: powi is missing because its libm implementation is not efficient
     fn powf(self, n: Self) -> Self => powf;
-    fn round(self) -> Self => roundf;
     fn sin_cos(self) -> (Self, Self) => sincosf;
     fn sqrt(self) -> Self => sqrtf;
 }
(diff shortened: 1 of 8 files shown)


color/src/floatfuncs.rs
```diff
@@ -43,18 +35,13 @@ macro_rules! define_float_funcs {
...
     fn atan2(self, other: Self) -> Self => atan2f;
     fn cbrt(self) -> Self => cbrtf;
     fn ceil(self) -> Self => ceilf;
-    // This is not needed once the MSRV is 1.84 or later.
-    fn copysign(self, sign: Self) -> Self => copysignf;
     fn floor(self) -> Self => floorf;
     fn hypot(self, other: Self) -> Self => hypotf;
     // Note: powi is missing because its libm implementation is not efficient
     fn powf(self, n: Self) -> Self => powf;
-    fn round(self) -> Self => roundf;
```
DJMcNab, line 57: This change needs sanity-checking; the current code never uses the `round()` function, except in a test.

waywardmonkeys:
This will require that Peniko bump MSRV from 1.85 to 1.86, but looking at what depends publicly on that, it doesn't seem like that's an issue. I don't see any big users of `color` that have a lower MSRV either although I didn't look exhaustively.

waywardmonkeys (approved):

xStrom (approved):

tomcur (approved):

color/src/floatfuncs.rs
tomcur: Removing it looks good to me.

PR #219, opened 2026-05-06
Add `DynamicColor::new` constructor for convenience

tomcur:


waywardmonkeys:
Maybe use this where a `DynamicColor` is constructed in parsing?

tomcur:
Ah, I now added it to the `from_alpha_color` constructor (missed it, but it's just below!). The parser uses its own `color_from_components` which builds the `Missing` flags

DJMcNab (approved):

PR #224, opened 2026-07-17
Add `Chromaticity::with_luminance` to convert chromaticities to colors

tomcur:
Chromaticity plus a luminance encodes a color. Mathematically the most straightforward coming from `xy` chromaticity is going to XYZ. As all our color spaces are relative to some reference white, this chooses to represent it as `XyzD65`, i.e., with a reference white of D65 (as that's the most common reference white).

The docs note to use `ColorSpace::convert_absolute` to convert this as an absolute color to another color space.

PR #225, opened 2026-07-17
Add Kelvin-to-chromaticity conversion

waywardmonkeys:
Add `Chromaticity::try_from_kelvin` using the Kang et al. Planckian-locus approximation over its documented range.

Returning a `Chromaticity` keeps color temperature independent of luminance, gamut mapping, and color-space adaptation. Callers can combine  it with `Chromaticity::with_luminance` and  `ColorSpace::convert_absolute` to configure linear-sRGB lighting without  changing the intended emitted color.


waywardmonkeys:
This is stacked on top of #224 which will have to land first.

